PR #1, opened 2018-10-03
Take `DEBUG` from config and apply it to ApolloServer

hedvigoscar:
This makes it not leak stack traces, which is good.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -4,6 +4,5 @@ const PLAYGROUND_ENABLED =
   (process.env.PLAYGROUND_ENABLED &&
     process.env.PLAYGROUND_ENABLED === 'true') ||
   false
-const NODE_ENV = process.env.NODE_ENV || 'development'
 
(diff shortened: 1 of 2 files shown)


src/config.ts
```diff
@@ -4,5 +4,6 @@ const PLAYGROUND_ENABLED =
   (process.env.PLAYGROUND_ENABLED &&
     process.env.PLAYGROUND_ENABLED === 'true') ||
   false
+const DEBUG = (process.env.DEBUG && process.env.DEBUG === 'true') || false
```
palmfjord: I don't like using the word `DEBUG`, considering its kinda claimed by http://npmjs.com/package/debug

hedvigoscar: Good point. What do you think about `APOLLO_DEBUG` or something similar?

palmfjord: Sure! 🙌🚀

hedvigoscar:
Made irrelevant by just using NODE_ENV='production' instead

PR #2, opened 2018-10-04
Sync schema on boot

hedvigoscar:


src/schema.ts
```diff
@@ -8,27 +8,41 @@ import {
...
-import translationSchema = require('./external-schemas/translations.json')
 import { resolvers } from './resolvers'
 import { typeDefs } from './typeDefs'
 
-const translationsLink = createHttpLink({
-  uri: 'https://api-euwest.graphcms.com/v1/cjmawd9hw036a01cuzmjhplka/master',
-  fetch: fetch as any,
-})
+const makeSchema = async () => {
+  const translationSchema = await fetch(
```
Iamsamwhoami: actually this is part of `graphql-tools` you could do:

```
introspectSchema(link).then((schema) => {
  // use the schema
});
```

hedvigoscar: :tada:

PR #4, opened 2018-10-08
Create Offer Mutation

hedvigoscar:
This PR grew a little bit out of hand, sorry 😞 

It basically contains:

- Create Offer Mutation
- Session management mutations
- Redis infrastructure
- Schema type generation infrastructure

If you think it's unmanageable in size I can split it down - just say the word :+1:


package.json
```diff
@@ -5,26 +5,35 @@
     "start": "node .",
     "dev": "ts-node src/index.ts",
     "build": "tsc",
-    "watch": "nodemon -e ts -w ./src -x yarn dev",
-    "lint": "tslint **/*.ts"
+    "watch": "concurrently \"nodemon -e ts -w ./src -x yarn dev\" \"nodemon -e graphqls -w ./src -x yarn generate:schema-types\"",
```
palmfjord: hoo tice! maybe we could give the commands some names though? So `concurrently -n \"ts-node,grahpqls\" ...`

src/api/index.ts
```diff
@@ -1,24 +1,72 @@
...
     method: 'POST',
-    headers: { Authorization: `Bearer ${token}` },
+    headers: headers(token),
+  })
+}
+
+const createProduct = (baseUrl: string) => async (token: string, body: any) => {
+  const data = await fetch(`${baseUrl}/insurance/createProductWeb`, {
+    method: 'POST',
+    headers: { ...headers(token), 'Content-Type': 'application/json' },
```
palmfjord: shouldn't content-type always be set to `application/json`? maybe we could move it to the `headers`function?

src/features/insurance/index.ts
```diff
@@ -0,0 +1,27 @@
...
+import * as config from '../../config'
+import { Insurance } from '../../typings/generated-graphql-types'
+
+const loadInsurance = async (token: string): Promise<Insurance> => {
+  const [insuranceObject, user] = await Promise.all([
+    getInsurance(config.BASE_URL)(token),
+    getUser(config.BASE_URL)(token),
+  ])
+  return {
+    insuredAtOtherCompany: insuranceObject.insuredAtOtherCompany,
```
palmfjord: even though it gets verbose i like this explicit remapping tbh

src/index.ts
```diff
@@ -13,16 +16,38 @@ makeSchema().then((schema) => {
...
   server.applyMiddleware({ app })
 
-  app.listen(
-    { port: config.PORT },
-    () => console.log(`Server listening at http://localhost:${config.PORT}`), // tslint:disable-line no-console
-  )
+  const ws = createServer(app.callback())
+
+  ws.listen(config.PORT, () => {
+    console.log(`Server listening at http://localhost:${config.PORT}`) // tslint:disable-line no-console
```
palmfjord: Actually, coming back to this - i think we should have some proper logging instead of console.log. Both because console.log isn't async meaning it blocks the thread (even though its for a very tiny amount of time) and because its nice to get things like timestamps n stuff

src/index.ts
```diff
@@ -13,16 +16,38 @@ makeSchema().then((schema) => {
...
+    new SubscriptionServer( // tslint:disable-line no-unused-expression
+      {
+        execute,
+        subscribe,
+        schema,
+        onConnect: (connectionParams: any) => {
+          const token = connectionParams.Authorization
+          return {
+            getToken: () => {
+              if (!token) {
```
palmfjord: 👌

palmfjord: Or... actually. Don't we need to validate the token anywhere in some way? hehe

src/schema.graphqls
```diff
@@ -0,0 +1,87 @@
...
+  insuredAtOtherCompany: Boolean
+  presaleInformationUrl: String
+  policyUrl: String
+  currentInsurerName: String
+
+  perilCategories: [PerilCategory]
+}
+
+type PerilCategory {
+  title: String
```
palmfjord: i feel like at least some of these props have to be required dont they?

src/api/index.ts
hedvigoscar: It isn't strictly speaking necessary when you're making `GET`'s though, given that you're not providing data to the receiver.

src/index.ts
hedvigoscar: We can if we want. We can also let `api-gateway` do that for us for now :wink:

src/schema.graphqls
hedvigoscar: There is one! Status will always return something. The other ones are not guaranteed to be set.

src/api/index.ts
palmfjord: Hm far enough. I guess we _should_ have an `Accept` header though?

src/index.ts
palmfjord: a-ha! As long as its validated :D carry on 👜

src/api/index.ts
```diff
@@ -1,24 +1,72 @@
...
+  activeFrom: string
+  categories: PerilCategory[]
+  insuredAtOtherCompany: boolean
+  personsInHousehold: number
+  presaleInformationUrl: string
+  currentInsurerName: string
+  policyUrl: string
+}
+
+interface UserDto {
```
Iamsamwhoami: Just User? Dto feels like a backend term.

src/features/offer/index.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+const offer: SubscriptionToOfferResolver = {
+  subscribe: withFilter(
+    () => pubsub.asyncIterator<OfferEvent>('OFFER_CREATED'),
+    (
+      payload: { offerCreated: OfferEvent },
+      variables: SubscriptionToOfferArgs,
+    ) => {
+      return payload.offerCreated.insuranceId === variables.insuranceId
```
Iamsamwhoami: shouldn't we also check that the user id is the same?

src/schema.ts
```diff
@@ -30,9 +37,9 @@ const makeSchema = async () => {
     [new FilterRootFields((_, name) => name === 'languages')],
   )
 
-  const localSchema = makeExecutableSchema({
-    typeDefs: gql(typeDefs),
-    resolvers,
+  const localSchema = makeExecutableSchema<Context>({
+    typeDefs,
+    resolvers: resolvers as IResolvers<any, Context>, // Not a big fan of this cast, but it will do for now
```
Iamsamwhoami: no comments remember ;)

hedvigoscar: Fine 😞

src/features/offer/index.ts
hedvigoscar: This is actually being redone in a future ticket! We can treat this as placeholder for the upcoming rework of this code.

Iamsamwhoami: oki!

src/api/index.ts
hedvigoscar: I had an idea of reserving none-DTO interfaces for proper GraphQL types (of which User has none).

src/api/index.ts
hedvigoscar: Yes! Fixing

src/middlewares/logger.ts
```diff
@@ -1,15 +1,6 @@
...
-} from 'typescript-logging'
+import { factory } from '../utils/log'
 
-const options = new LoggerFactoryOptions().addLogGroupRule(
-  new LogGroupRule(/.+/, LogLevel.fromString('info')),
-)
-const factory = LFService.createLoggerFactory(options)
 const logger = factory.getLogger('koa')
 
 const loggingMiddleware = async (
```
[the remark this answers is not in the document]
hedvigoscar: I like this idea and I want to do it, but I'm going to scope it out for now and make a later PR to fix it.

PR #7, opened 2018-10-09
Complete the Offer Status subscription

hedvigoscar:


hedvigoscar:
@HedvigInsurance/clients

src/features/offer/index.ts
```diff
@@ -36,24 +35,24 @@ const createOffer: MutationToCreateOfferResolver = async (
 }
 
 const getInsuranceByOfferSuccessEvent: OfferEventToInsuranceResolver = async (
-  _parent,
+  parent,
   _args,
   { getToken, headers },
 ) => {
+  if (parent.status === 'FAIL') {
+    throw new ApolloError('Failed to create offer')
```
palmfjord: Maybe we should add a because?

hedvigoscar: There's actually no valid cases where you should not be getting an offer, only if there's something really strange (we send `status: 'fail'` when Hibernate fails to commit the offer to the database). 🤷‍♂️

PR #9, opened 2018-10-10
Sign Offer Mutation

hedvigoscar:


src/api/index.ts
```diff
@@ -39,6 +74,7 @@ const register = (baseUrl: string, headers: ForwardHeaders) => async () => {
     method: 'POST',
     headers: headers as any,
   })
+  checkStatus(data)
```
palmfjord: i really don't like doing these types of checks every time we want to call the api. Can't we just abstract the fetch into something we can reuse?

PR #10, opened 2018-10-10
Subscribe to Offer Sign Status

hedvigoscar:
I may have messed up the branch name a bit. But that's ok 🙂

This branch is based on `feature/sign-offer-mutation` (#9) - review that before this 🙏


src/api/index.ts
```diff
@@ -81,7 +120,38 @@ const createProduct = (baseUrl: string, headers: ForwardHeaders) => async (
     },
     body: JSON.stringify(body),
   })
+  checkStatus(data)
```
palmfjord: i really don't like doing these types of checks every time we want to call the api. Can't we just abstract the fetch into something we can reuse?

palmfjord: woops, i'll leave the same comment in #9 hehe

src/api/index.ts
```diff
@@ -57,92 +58,131 @@ interface CreateProductDto {
...
+  },
+) => Promise<Response>
+
+const callApi: CallApi = async (url, options = {}) => {
+  const { mergeOptions, token, validateStatus = checkStatus } = options
+  const headers: { [key: string]: string } = {
+    Accept: 'application/json',
+  }
+  if (mergeOptions!.method === 'POST') {
+    headers['Content-Type'] = 'application/json'
```
palmfjord: i really don't like these kind of mutations on an object, but lets leave it for now, its quite isolated

src/api/index.ts
```diff
@@ -57,92 +58,131 @@ interface CreateProductDto {
...
-
-const register = (baseUrl: string, headers: ForwardHeaders) => async () => {
-  const data = await fetch(`${baseUrl}/helloHedvig`, {
-    method: 'POST',
-    headers: headers as any,
+const register = async (headers: ForwardHeaders) => {
+  const data = await callApi('/helloHedvig', {
+    mergeOptions: {
+      method: 'POST',
+      headers: headers as any,
```
palmfjord: why u no

typecast properly

src/api/index.ts
hedvigoscar: I agree actually, this is just the least bad way I've found to represent this sort of thing 😕 I'm open for suggestions if you have alternative ways to represent it!

PR #14, opened 2018-10-12
Log responsibly

hedvigoscar:
- We now log out the inner errors in the graphql error


hedvigoscar:
Like keeping an ID for `createSession => subscribe offer => createOffer`? That's a good idea! I think we'll have to push to after launch.

PR #17, opened 2018-10-13
Add the possibility to upload files

Iamsamwhoami:
This PR makes it possible to upload files into S3 🎉


src/resolvers/uploadFile.ts
```diff
@@ -0,0 +1,41 @@
+/// <reference path="../typings/cloudinary.d.ts" />
+import { UserInputError } from 'apollo-server-koa'
+import * as cloudinary from 'cloudinary'
+
+import {
+  File,
+  MutationToUploadFileResolver,
+} from '../typings/generated-graphql-types'
+
+cloudinary.config({
```
hedvigoscar: Can we use `src/config.ts` as a single source of configuration variables? I think it's a nice convenient way to keep track of all configuration the application takes in, provide reasonable defaults if applicable and ensure type correctness. 🙂

hedvigoscar: Also, if applicable we might want to add some entries to the example env file 👌

src/typings/generated-graphql-types.ts
```diff
@@ -119,6 +120,12 @@ export interface SignInput {
   email: string;
 }
 
+export type Upload = any;
```
hedvigoscar: We may want to contribute an improvement for this one in the future 🤷‍♂️

PR #21, opened 2018-10-15
Expose member info needed for offer page

hedvigoscar:
- New root query: member
- Added postal number to insurance

### DIFF
--- a/src/features/insurance/index.ts
+++ b/src/features/insurance/index.ts
@@ -18,6 +18,7 @@ const loadInsurance = async (
     presaleInformationUrl: insuranceObject.presaleInformationUrl,
     monthlyCost: insuranceObject.currentTotalPrice,
     address: insuranceObject.addressStreet,
+    postalNumber: insuranceObject.zipCode,
     certificateUrl: insuranceObject.certificateUrl,
     status: insuranceObject.status,
     type: insuranceObject.insuranceType,
--- a/src/features/insurance/index.ts
+++ b/src/features/insurance/index.ts
@@ -6,24 +6,24 @@ const loadInsurance = async (
   token: string,
   headers: ForwardHeaders,
 ): Promise<Insurance> => {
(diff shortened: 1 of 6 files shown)


src/features/insurance/index.ts
```diff
@@ -18,6 +18,7 @@ const loadInsurance = async (
     presaleInformationUrl: insuranceObject.presaleInformationUrl,
     monthlyCost: insuranceObject.currentTotalPrice,
     address: insuranceObject.addressStreet,
+    postalNumber: insuranceObject.zipCode,
```
palmfjord: can we decide on `postalCode`, `postalNumber` or `zipCode` maybe 😅 i'm starting to get confused here

src/resolvers/member.ts
```diff
@@ -0,0 +1,18 @@
+import { getUser } from '../api'
+import { QueryToMemberResolver } from '../typings/generated-graphql-types'
+
+const member: QueryToMemberResolver = async (
+  _parent,
+  _args,
+  { getToken, headers },
+) => {
+  const token = await getToken()
+  const memberObj = await getUser(token, headers)
```
palmfjord: `member`? we know it's an object already right?

src/features/insurance/index.ts
hedvigoscar: I'm trying to standardize on `postalNumber`, but we're exposing `zipCode` from backend in a lot of places.

src/resolvers/member.ts
hedvigoscar: `member` shadows the function name 😞 I know it's ugly and I'm open to better names

src/features/insurance/index.ts
palmfjord: ok, coolio! carry on 👜

src/resolvers/member.ts
palmfjord: ): `memberResponse`?

PR #33, opened 2018-10-19
fix mistake 🙈

Iamsamwhoami:
`memberId` should have been `memberid` 🙅‍♂️


Iamsamwhoami:
@palmfjord the metadata object isn't typed as it's an arbitrary dataset 😭

PR #42, opened 2018-10-25
Allow 404 from sign endpoint and allow using real ip

palmfjord:


src/context.ts
```diff
@@ -57,6 +59,9 @@ const getWebSocketContext = (
   return {
     getToken,
     headers,
+    remoteIp:
+      headers['X-Forwarded-For'] ||
+      (context.request.connection.address() as string),
```
hedvigoscar: Good catch! I did not know about this when I set these things up but it's clearly way better 👌

PR #55, opened 2018-10-31
Add things to tracking entity and such.

hedvigoscar:


hedvigoscar:
This code was less complete than I thought. Do not approve. Do not merge. 🙁

hedvigoscar:
Now it's a bit better @HedvigInsurance/clients

src/resolvers/createSession.ts
```diff
@@ -1,12 +1,36 @@
...
+    campaignPromise = registerCampaign(token, headers, {
+      utmSource: campaign.source,
+      utmMedium: campaign.medium,
+      utmContent: campaign.content ? [campaign.content] : undefined,
+      utmCampaign: campaign.name,
+      utmTerm: campaign.term ? [campaign.term] : undefined,
+    })
+  }
+
+  let trackingIdPromise
```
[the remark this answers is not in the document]
hedvigoscar: What I really wanted to express was that block of code, actually! :+1:

src/api/index.ts
```diff
@@ -287,6 +295,34 @@ const registerDirectDebit = async (
   return data.json()
 }
 
+const registerCampaign = async (
```
palmfjord: no need for this to be async anymore

hedvigoscar:
```suggestion
const registerCampaign = (
```

src/api/index.ts
```diff
@@ -287,6 +295,34 @@ const registerDirectDebit = async (
...
+  callApi('/hedvig/register_campaign', {
+    mergeOptions: {
+      headers: (headers as any) as RequestInit['headers'],
+      method: 'POST',
+      body: JSON.stringify(body),
+    },
+    token,
+  })
+
+const assignTrackingId = async (
```
hedvigoscar:
```suggestion
const assignTrackingId = (
```

PR #59, opened 2018-11-01
Log errors for both regular and subscription errors. and sentry.

hedvigoscar:
Sometimes you make me a little disappointed, apollo.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -17,7 +17,7 @@ import * as Sentry from '@sentry/node'
 Sentry.init({
   dsn: config.SENTRY_DSN,
   enabled: Boolean(config.SENTRY_DSN),
-  environment: process.env.NODE_ENV,
+  environment: config.SENTRY_ENV,
 })
 
 const logger = factory.getLogger('index')
--- a/src/index.ts
+++ b/src/index.ts
@@ -14,6 +14,7 @@ import { makeSchema } from './schema'
 import { factory } from './utils/log'
 
 import * as Sentry from '@sentry/node'
+import { getInnerErrorsFromCombinedError } from './utils/graphql-error'
 Sentry.init({
   dsn: config.SENTRY_DSN,
   enabled: Boolean(config.SENTRY_DSN),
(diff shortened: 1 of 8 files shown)


src/index.ts
```diff
@@ -13,8 +13,33 @@ import { loggingMiddleware } from './middlewares/logger'
 import { makeSchema } from './schema'
 import { factory } from './utils/log'
 
+import * as Sentry from '@sentry/node'
+Sentry.init({
+  dsn: config.SENTRY_DSN,
+  enabled: Boolean(config.SENTRY_DSN),
+  environment: process.env.NODE_ENV,
```
palmfjord: imo this should be a separate env var. we probably want different environments for `staging` and `production` for example, but `NODE_ENV` _should_ be `production` for both

src/index.ts
```diff
@@ -13,8 +13,33 @@ import { loggingMiddleware } from './middlewares/logger'
...
+})
+
 const logger = factory.getLogger('index')
 
+const handleError = (error: GraphQLError): void => {
+  logger.error(
+    'Uncaught error in GraphQL. Original error: ',
+    error.originalError,
+  )
+  Sentry.captureException(error)
```
palmfjord: would it be possible to do some more lovin with scoping? I think this'll kinda catch all errors and breadcrumbs from the beginning of time

hedvigoscar: I'm not sure that I follow 🙁 I'd love to fix this however!

palmfjord: I'm thinking something like this https://github.com/HedvigInsurance/web-onboarding/blob/master/src/server/middleware/enhancers.ts#L88-L100

hedvigoscar: From what I can tell Apollo Server eats the error before it comes back to Koa, making a regular Koa Middleware not the ideal approach (?). Nor does Apollo Server offer any nice places to perform the 'wrapping' we're looking for in an around-style middleware.

Apollo is starting to make me a little bit sad 🙁

hedvigoscar: New implementation - Ping! @palmfjord

src/schema.ts
```diff
@@ -49,7 +51,8 @@ const makeSchema = async () => {
     schemas: [transformedTranslationSchema, localSchema],
   })
 
-  return schema
+  // return schema
```
palmfjord: 🙈

hedvigoscar: 🙈 consider using the suggestions feature for this! Perfect use case I think 🙂

PR #108, opened 2018-11-15
Add chat

Iamsamwhoami:
This PR adds the chat logic to giraffe:

- Polling
- Getting chat state
- Getting chat messages
- Responding with text

Some stuff left:
- Transform all messages
- Respond with multiple select


src/api/index.ts
```diff
@@ -85,6 +93,27 @@ interface DirectDebitOrderInfoDto {
...
+
+export interface MessageDto {
+  id: string
+  globalId: string
+  header: MessageHeaderDto
+  body: MessageBody
+}
+
+export interface ChatDto {
+  state: ChatState
```
hedvigoscar: While this may hold true for now it introduces some 'sneaky' coupling between our GraphQL types and our backend types. Maybe this is not desirable?

src/schema.graphqls
```diff
@@ -173,8 +347,12 @@ scalar URL
 scalar UUID
 
 type File {
-  """signedUrl is valid for 30 minutes after upload, don't hang on to this."""
+  """
```
```suggestion
  # signedUrl is valid for 30 minutes after upload, don't hang on to this.
```

src/resolvers/messages.ts
```diff
@@ -0,0 +1,106 @@
...
+  _root,
+  _args,
+  { getToken, headers },
+) => {
+  const token = getToken()
+  const chat = await getChat(token, headers)
+  return transformMessages(chat.messages)
+}
+
+export const subscribeToMessage: SubscriptionToMessageResolver = {
```
hedvigoscar: I'm very much in favor of us moving the subscriptions to redis and letting bot-service publish - I dont think we should deploy the app with this in place.

But lets not let that prevent us from merging to test, I think

src/schema.graphqls
```diff
@@ -19,11 +22,182 @@ type Mutation {
...
+  webUrl: String
+}
+
+type ChatState {
+  ongoingClaim: Boolean!
+  showOfferScreen: Boolean!
+  onboardingDone: Boolean!
+}
+
+type ChatResponse {
```
hedvigoscar: This type is equal to `Message`, right?

src/resolvers/currentChatResponse.ts
```diff
@@ -0,0 +1,56 @@
...
+import {
+  ChatResponse,
+  QueryToCurrentChatResponseResolver,
+  SubscriptionToCurrentChatResponseResolver,
+} from '../typings/generated-graphql-types'
+
+import { subscribeToChat } from '../features/chat/chatSubscription'
+import { transformMessage } from '../features/chat/transform'
+
+export const currentChatResponse: QueryToCurrentChatResponseResolver = async (
```
hedvigoscar: I'm not entirely sure that I understand the use-case for this resolver? 🤔

src/schema.graphqls
Iamsamwhoami: @hedvigoscar that doesn't work then it doesn't show up in the docs 🙈

src/api/index.ts
Iamsamwhoami: As they are exactly the same right now I guess it's fine? 🤔

src/schema.graphqls
Iamsamwhoami: I wanted to dissociate them due to the docs being clearer if they are, but yeah it's basically a message, makes sense?

src/resolvers/currentChatResponse.ts
Iamsamwhoami: Get the first message, which is the message that a user can respond to, in order to not have that logic in the frontend 😄

src/resolvers/messages.ts
Iamsamwhoami: Yeap

src/schema.graphqls
hedvigoscar: 😞 🔫

src/resolvers/currentChatResponse.ts
hedvigoscar: Makes sense! I guess the name is slightly unclear though 🤔

src/schema.graphqls
hedvigoscar: Perhaps we could re-shape the data to be a little bit more sane. Also maybe not. 🤔 I'm not sure

Iamsamwhoami: The data structure is hell 🔥

src/resolvers/currentChatResponse.ts
Iamsamwhoami: What do you think is a better name? :)

PR #117, opened 2018-11-20
Add file response

Iamsamwhoami:
This makes it possible to send a file response to Hedvig.

This little thing is cherry-picked from: https://github.com/HedvigInsurance/giraffe/pull/108


src/api/index.ts
```diff
@@ -295,6 +326,57 @@ const registerDirectDebit = async (
...
+  )
+
+  if (!responseMessage) {
+    throw new Error("Tried to respond to a message that doesn't exist")
+  }
+
+  const responseMessageWithFile = {
+    ...responseMessage,
+    body: {
+      ...responseMessage.body,
```
palmfjord:
```ts
...responseMessage.body,
...responseInput.body,
```
?

Iamsamwhoami:
@palmfjord fixed 👀

src/api/index.ts
```diff
@@ -120,13 +145,19 @@ const callApi: CallApi = async (url, options = {}) => {
   }
 
   const res = await fetch(`${config.BASE_URL}${url}`, requestOptions)
-  validateStatus(res)
+  await validateStatus(res)
```
palmfjord: now that we're `await`ing this, shouldn't the rest of the `validateStatus` functions also be async? like this guy for example:  https://github.com/HedvigInsurance/giraffe/pull/117/files#diff-55e1fcd9400087f22bb1f21cfaf39385R250

Iamsamwhoami: 🙃 didn't realize, removed!

PR #147, opened 2019-01-29
Added mutations for member-phonenumber and email

johantjellden:


src/resolvers/member.ts
```diff
@@ -17,4 +17,28 @@ const member: QueryToMemberResolver = async (
...
+const updateEmail: MutationToUpdateEmailResolver = async (
+  _root,
+  {input},
+  { getToken, headers },
+) => {
+  const token = getToken()
+  await postEmail(token, headers, {
+    newEmail: input
+  })
+  return true
```
johantjellden: What about this? Should I return something smarter here?

Iamsamwhoami: I think it's fine, the `await` will throw if some error occur else the function will return `true`.

johantjellden: Awesome

PR #196, opened 2019-02-28
add new mutation to set the active to

meletios:


src/schema.graphqls
```diff
@@ -31,6 +31,7 @@ type Mutation {
   editLastResponse: Boolean!
   updateEmail(input: String!): Member!
   updatePhoneNumber(input: String!): Member!
+  activateAtDate(id: UUID!, activationDate: String! ): Boolean!
```
Iamsamwhoami: Shouldn't this be `activateOnDate`? 🤔 https://www.englishclub.com/grammar/prepositions-at-in-on-time.htm

Iamsamwhoami: And I think you can skip `date` in the mutation, that is implied.

Iamsamwhoami: also, maybe rename `id` to `insuranceId` to make it clearer what the ID is for

Iamsamwhoami: And `activationDate` should be a `Date!` not a `String!` I guess

Iamsamwhoami: so `activateOn(date: Date!, insuranceId: UUID!)`

gustafguner: Yeah, you're right. Makes more sense 👍

gustafguner:
This PR is dependent on PR #138 in product-pricing: https://github.com/HedvigInsurance/product-pricing/pull/138. 

Giraffe now uses the new endpoint `/insurance/{memberId}/requestStartDate` instead.

PR #263, opened 2019-04-29
Chat triggers

hedvigoscar:


src/api/index.ts
```diff
@@ -628,6 +628,24 @@ const registerPushToken = (
     token,
   })
 
+const triggerFreeTextChat = (token: string, headers: ForwardHeaders) =>
+  callApi('/v2/app/fabTrigger/CHAT', {
```
palmfjord: is this case sensitive? If so i'm skeptical. Otherwise awesome

hedvigoscar: I mean, the RFC says that case sensitivity is not a thing for URLs, but I'm not sure how Spring Boot actually treats it 🤔

hedvigoscar: After checking: it is case sensitive, this is an imposed constraint by Spring Boot.

PR #317, opened 2019-06-03
Include product pricing schema

FredrikAreschoug:


.env.example
```diff
@@ -8,5 +8,6 @@ AWS_SECRET=your_secret
 AWS_S3_BUCKET=your_bucket
 DONT_PANIC_ENDPOINT=http://localhost:4242/grahpql
 PAYMENT_SERVICE_GRAPHQL_ENDPOINT=http://localhost:2424/graphql
+PRODUCT_PRICING_SERVICE_GRAPHQL_ENDPOINT=http://localhost:2525/graphql
```
hedvigoscar: Does product-pricing usually run on 2525?

FredrikAreschoug: Good catch apparently it usually uses 4085 :)

PR #386, opened 2019-07-11
Claim uploads, in one go.

hedvigoscar:
This PR requires a new environment variable for
both dev and prod:
AWS_CLAIMS_S3_BUCKET

Unfortunately, no messages are emitted on the subscription after performing the mutation.

Stupid bot-service. 😞


src/resolvers/claims.ts
```diff
@@ -0,0 +1,48 @@
...
+        logger.error('Got error when attempting to upload', err)
+        reject(err)
+        return
+      }
+
+      resolve(data.Location)
+    })
+  })
+
+  return setChatAudioResponse(token, headers, {
```
Iamsamwhoami: Uploading a claim file shouldn't be the same as sending a message IMO

hedvigoscar: This is mostly what the client does right now, but this does it in a better way. I agree that it should not really be sending messages in the first place, but it is what it is.

Iamsamwhoami: Yeah, but it's better that the client does it on its own, decoupling is good :)

PR #410, opened 2019-07-25
start bank id sign from app

FredrikAreschoug:


src/features/offer/sign.ts
```diff
@@ -74,9 +75,29 @@ const subscribeToSignStatus: SubscriptionToSignStatusResolver = {
...
+  _args,
+  { getToken, headers, remoteIp },
+) => {
+  const token = getToken()
+  const signDetailsResult = await signDetails(token)
+
+  const websignResult = await websign(token, headers, {
+    email: signDetailsResult.email,
+    ssn: signDetailsResult.personalNumber,
+    ipAddress: remoteIp,
```
hedvigoscar: We should validate that we have gotten remoteIp I think, like in the other mutation 🙂

src/schema.graphqls
```diff
@@ -451,6 +452,10 @@ type BankIdAuthResponse {
   autoStartToken: String!
 }
 
+type BankIdSignResponsResponse {
```
```suggestion
type BankIdSignResponse {
```

PR #439, opened 2019-08-07
Add renewal certificates+dates to insurance schema

vonElfvin:
Also added added startup logging for AppContentService


src/schema.graphqls
```diff
@@ -373,6 +373,8 @@ type Insurance {
     reason: "No longer supported"
   )
   arrangedPerilCategories: ArrangedPerilCategories!
+  renewalCertificateUrl: String
```
```suggestion
  renewal: Renewal
```

and then:

```
type Renewal {
  url: Url!
  date: LocalDate!
}
```

PR #465, opened 2019-08-20
Add more information about previous insurer and if we can switch or not

Iamsamwhoami:
The logic for identifying which providers we can switch from was in the app repo before, and it doesn't make too much sense to have it in two separate code bases going forward (android & ios), so moving the logic to here instead.

poke @hedvigoscar @FredrikAreschoug you might like this

### DIFF
--- a/src/schema.graphqls
+++ b/src/schema.graphqls
@@ -383,7 +383,7 @@ type Insurance {
 
 type PreviousInsurer {
   displayName: String
-  identifier: String!
+  id: ID!
   switchable: Boolean!
 }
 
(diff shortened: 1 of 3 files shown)


src/schema.graphqls
```diff
@@ -374,6 +378,13 @@ type Insurance {
   )
   arrangedPerilCategories: ArrangedPerilCategories!
   renewal: Renewal
+  previousInsurer: PreviousInsurer
+}
+
+type PreviousInsurer {
+  displayName: String
+  identifier: String!
```
```suggestion
  identifier: ID!
```

Iamsamwhoami: @hedvigoscar makes sense, updated!

hedvigoscar: 👏

PR #504, opened 2019-09-25
[WIP] Enable house

FredrikAreschoug:
Nothing super fancy, but please have an extra look at the schema

### DIFF
--- a/src/schema.graphqls
+++ b/src/schema.graphqls
@@ -410,6 +410,25 @@ type Insurance {
   arrangedPerilCategories: ArrangedPerilCategories!
   renewal: Renewal
   previousInsurer: PreviousInsurer
+
+  ancillaryArea: Int
+  yearOfConstruction: Int
+  numberOfBathrooms: Int
+  extraBuildings: [ExtraBuilding!]
+  isSubleted: Boolean
+}
+
+type ExtraBuilding {
+  type: ExtraBuildingType
+  area: Int!
+  hasWaterConnected: Boolean!
+}
+
+enum ExtraBuildingType {
+  GARAGE
+  ATTEFALL
+  FRIGGEBOD
+  OTHER
 }
 
 type PreviousInsurer {
--- a/src/schema.graphqls
+++ b/src/schema.graphqls
@@ -410,7 +410,6 @@ type Insurance {
   arrangedPerilCategories: ArrangedPerilCategories!
   renewal: Renewal
   previousInsurer: PreviousInsurer
-
   ancillaryArea: Int
   yearOfConstruction: Int
   numberOfBathrooms: Int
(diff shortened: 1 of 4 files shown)


src/schema.graphqls
```diff
@@ -410,6 +410,24 @@ type Insurance {
   arrangedPerilCategories: ArrangedPerilCategories!
   renewal: Renewal
   previousInsurer: PreviousInsurer
+  ancillaryArea: Int
```
hedvigoscar: These fields are only relevant if you have house insurance, could we make that clear in the type information? I'm generally not in favour of enum-style unions when we could use true unions

FredrikAreschoug: True I will look into creating a union from insurance the question is if this could be done without breaking the schema

hedvigoscar: I think this could be a valid candidate for a V2 if we can't do it without breaking the schema

Iamsamwhoami: yeah this can probably be solved neatly with, unions and interfaces, maybe it's time to add an insuranceV2 and make it an array? which can contain either house or apartment insurances.

have a look at: https://github.com/HedvigInsurance/giraffe/blob/master/src/schema.graphqls#L90 for inspo

src/schema.graphqls
```diff
@@ -410,6 +410,24 @@ type Insurance {
...
+
+type ExtraBuilding {
+  type: ExtraBuildingType
+  area: Int!
+  hasWaterConnected: Boolean!
+}
+
+enum ExtraBuildingType {
+  GARAGE
+  ATTEFALL
```
Iamsamwhoami: I have no idea what I feel about these, can we abstract this very Swedish-market specific thing to something else? or perhaps make it more explicit somehow.

FredrikAreschoug: I'm guessing these types will be very market specific, should we really translate them?

FredrikAreschoug: Yeah we should handle stuff like this better. Maybe calling them `SwedishExtraBuildingType` is enough. Maybe even the different insurance types should also define which market it belongs too

src/schema.graphqls
hedvigoscar: We can save this until we have the new product-pricing

src/schema.graphqls
hedvigoscar: This however I think we should not save until the next product-pricing

FredrikAreschoug:
Okay @hedvigoscar what do you think is best calling it `SwedishExtraBuildingType` or do you have another suggestion, unifying them for different markets will be really hard and I don't have the knowledge of doing that.

src/schema.graphqls
FredrikAreschoug: Okay @hedvigoscar @Iamsamwhoami  what do you think is best calling it SwedishExtraBuildingType or do you have another suggestion, unifying them for different markets will be really hard and I don't have the knowledge of doing that.

PR #552, opened 2019-10-23
forward x-attribution-code header

FredrikAreschoug:
Forward attribution code to be able to store partner when singing up with dreams for example


hedvigoscar:
I'm lacking a bit of context for this PR, what is this for? 🙂

PR #592, opened 2019-11-26
stitch in underwriter

FredrikAreschoug:


FredrikAreschoug:
Yeah, maybe that's a discussion we can have. It feels like it's going to be a pain for testing. But can we find a solution for that I'm game!

PR #691, opened 2020-03-02
Add features concept

sampettersson:


hedvigoscar:
I dislike this of course, but let's go for it and kill it soon

PR #698, opened 2020-03-04
update market mutation 

akashpatl:
Country code should be posted as  ISO 3166 alpha-2

### DIFF
--- a/src/schema.graphqls
+++ b/src/schema.graphqls
@@ -47,7 +47,7 @@ type Mutation {
   norwegianBankIdAuth: NorwegianBankIdAuthResponse!
   registerBranchCampaign(campaign: CampaignInput!): Boolean
   updateLanguage(input: String!): Boolean! # input should be in the format as Accept-Language header
-  updateMarket(input: Market!): Boolean! # input should be in the format of  ISO 3166  alpha-2 country code
+  updateMarket(market: Market!): Member! # input should be in the format of  ISO 3166  alpha-2 country code
 }
 enum Market {
   SE
--- a/src/schema.graphqls
+++ b/src/schema.graphqls
@@ -47,9 +47,9 @@ type Mutation {
   norwegianBankIdAuth: NorwegianBankIdAuthResponse!
   registerBranchCampaign(campaign: CampaignInput!): Boolean
   updateLanguage(input: String!): Boolean! # input should be in the format as Accept-Language header
-  updateMarket(market: Market!): Member! # input should be in the format of  ISO 3166  alpha-2 country code
+  updatePickedLocale(pickedLocale: PickedLocale!): Member! # input should be in the format of  ISO 3166  alpha-2 country code
 }
-enum Market {
+enum PickedLocale {
   SE
   NO
 }
--- a/src/schema.graphqls
+++ b/src/schema.graphqls
@@ -46,7 +46,8 @@ type Mutation {
   swedishBankIdAuth: BankIdAuthResponse!
   norwegianBankIdAuth: NorwegianBankIdAuthResponse!
   registerBranchCampaign(campaign: CampaignInput!): Boolean
-  updateLanguage(input: String!): Boolean! # input should be in the format as Accept-Language header
+  updateMarket(input: String!): Boolean! # input should be in the format as Accept-Language header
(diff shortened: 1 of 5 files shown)


hedvigoscar:
I loke the concept of a draft! I have not used it before but I will from now on 👍

src/schema.graphqls
```diff
@@ -46,7 +46,8 @@ type Mutation {
   swedishBankIdAuth: BankIdAuthResponse!
   norwegianBankIdAuth: NorwegianBankIdAuthResponse!
   registerBranchCampaign(campaign: CampaignInput!): Boolean
-  updateLanguage(input: String!): Boolean! # input should be in the format as Accept-Language header
+  updateMarket(input: String!): Boolean! # input should be in the format as Accept-Language header
+  updateLanguage(input: String!): Boolean! # input should be in the format of  ISO 3166  alpha-2 country code
```
hedvigoscar: Are these comments mixed up? 🙂

src/schema.graphqls
```diff
@@ -46,7 +46,8 @@ type Mutation {
   swedishBankIdAuth: BankIdAuthResponse!
   norwegianBankIdAuth: NorwegianBankIdAuthResponse!
   registerBranchCampaign(campaign: CampaignInput!): Boolean
-  updateLanguage(input: String!): Boolean! # input should be in the format as Accept-Language header
+  updateMarket(input: String!): Boolean! # input should be in the format as Accept-Language header
```
hedvigoscar: How about we introduce a `Scalar` for the input here? I think it can be fitting as the input is constrained to a particular form of `String`, and with a `Scalar` we can automatically convert it to a fitting type if there is one 🙂

hedvigoscar: Also, should the Market be exposed on the `Member`, and if yes, should we return the updated member from this mutation? As it stands, the `Market`-property is now read-only, which is not very useful 🙁

src/schema.graphqls
akashpatl: Yes! Good catch!

src/schema.graphqls
akashpatl: For the first comment I actually changed it to an enum. Because that's how we handle it in the backend anyways. 

For the second one I don't really see the point for returning the member unless we need to do something with the returned member entity. So I'll leave it for now. :)

hedvigoscar: An enum is a good option here I think!

I disagree about the mutation return type, however - not returning the entity containing the updated data reduces the usefulness of the mutation and potentially introduces scenarios where consumers need to forcibly refresh their local data - complicating API consumption and introducing more points of failure. Additionally, if we change our minds later and want to have this functionality later, we will be forced to make a V2-mutation, as GraphQL mandates that your schema is fully backwards compatible.

Adding `Member` as the return type is fairly easy, we can simply call the API-endpoint at the end of the resolver function and call it a day, I think.

src/schema.graphqls
```diff
@@ -47,6 +47,11 @@ type Mutation {
   norwegianBankIdAuth: NorwegianBankIdAuthResponse!
   registerBranchCampaign(campaign: CampaignInput!): Boolean
   updateLanguage(input: String!): Boolean! # input should be in the format as Accept-Language header
+  updateMarket(input: Market!): Boolean! # input should be in the format of  ISO 3166  alpha-2 country code
```
sampettersson: Generally you do `input: UpdateMarketInput` where `UpdateMarketInput` would contain the field `market: Market`, this is for keeping backward compatibility if you ever need to add more fields to the input for whatever reason :)

akashpatl: 👍

src/schema.graphqls
akashpatl: 👍

hedvigoscar:
To be honest, I no longer understand this mutation. Is it only for analytics-purposes?

akashpatl:
> To be honest, I no longer understand this mutation. Is it only for analytics-purposes?

Yes to some extent. But my understanding was that you need this for the market picker to also work?

hedvigoscar:
Well, as we cant query the information, there's no way we can utilize it 🤔

PR #718, opened 2020-03-19
Remove auth from app-content-service

hedvigoscar:


hedvigoscar:
We actually have a proper use case for this now: It needs to be queried from the Market-Web in order to show perils on different landing pages.

Merging!

PR #720, opened 2020-03-20
added referrals to features

FredrikAreschoug:
Referrals in the feature list. I stole @sampettersson amazing idea of checking if the member has been referred to in product and pricing. So no more checks are needed. If the user gets referrals in features list show referrals otherwise don't :)

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -775,6 +775,22 @@ const signDetails = (
     token,
   }).then((res) => res.json())
 
+const isEligibleForReferrals = async (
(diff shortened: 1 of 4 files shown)


sampettersson:
Does `isEligibleForReferrals` always return true for SE? 😄

FredrikAreschoug:
yes `isEligibleForReferrals` logic is:
`member with swedish contract` -> true 
`member with norwegian contract, referred` -> true 
 `member with norwegian contract not referred` -> false

PR #730, opened 2020-03-26
changed enum type and also the return type to Boolean

akashpatl:


hedvigoscar:
I believe we discussed this before, that a more appropriate schema would return the updated view of the member post mutation?

Then again this is now a read-only property (arguably not a great thing), so I guess it's whatever. We're limited to never using this property in any frontend under this schema.

akashpatl:
Yes we have. My reasoning is that since we've concluded this discussion before I think we should just stick with that decision (even if it's not 100 % correct) and we can fix in the future should the need arise.   :)

hedvigoscar:
I'm still confused, was the conclusion not that we should return the member? Or has there been a separate conclusion in a forum I was not part of?

akashpatl:
returns Member now! :D

akashpatl:
@hedvigoscar Do we feel ready with this now? Member service is in staging so you can test this as well :)

hedvigoscar:
No, let's merge 🙂

PR #759, opened 2020-04-20
Forward them IPs

hedvigoscar:


src/schema.ts
```diff
@@ -112,6 +112,12 @@ const makeSchema = async () => {
       authorization: `Bearer ${previousContext.graphqlContext &&
         previousContext.graphqlContext.getToken &&
         previousContext.graphqlContext.getToken()}`,
+      ...(previousContext.graphqlContext &&
+        previousContext.graphqlContext.headers &&
+        previousContext.graphqlContext.headers && {
```
palmenhq, line 117: i think we have one `previousContext.graphqlContext.headers` too much or?

src/schema.ts
```diff
@@ -112,6 +112,12 @@ const makeSchema = async () => {
       authorization: `Bearer ${previousContext.graphqlContext &&
         previousContext.graphqlContext.getToken &&
         previousContext.graphqlContext.getToken()}`,
+      ...(previousContext.graphqlContext &&
```
palmenhq, line 115: what happens if we try spreading `undefined`? i think this maybe should be 
```
<the has headers check> ? {
  'x-forwarded-for': ...,
} : {}
```

its not pretty but it's darn safe

src/schema.ts
```diff
@@ -112,6 +112,12 @@ const makeSchema = async () => {
       authorization: `Bearer ${previousContext.graphqlContext &&
```
palmenhq, line 112: two lines up, shall we also rename this from `authorizationContextLink`?

src/schema.ts
hedvigoscar: spreading undefined is valid and results in nothing:

```
> {foo:'bar', ...(undefined)}
{ foo: 'bar' }
```

hedvigoscar: This actually holds true for most falsy values:
```
> {foo:'bar',...(false)}
{ foo: 'bar' }
> {foo:'bar',...(null)}
{ foo: 'bar' }
```

src/schema.ts
hedvigoscar: `authorizationAndIpLink`? Honestly I think the name is still kind of fine 🤔

palmenhq: still think its confusing haha, maybe it should be two separate links rather?

hedvigoscar: It is admittedly a hack. Two links might be the way to go. Having this functionality for all links would probably be desirable, so maybe the IP-one can serve as a base-link that the others join upon

PR #934, opened 2020-10-15
enable danish auth

FredrikAreschoug:
I totally forgot about giraffe but this should enable danish auth


hedvigoscar (approved):
Looks good, I suppose. Would it not be easier to just merge in GraphQL from the upstream service, though? 🤔

PR #949, opened 2020-10-28
Fix/cross schema stiching

vonElfvin:


palmenhq (approved):
Not sure i understand it 🤷‍♀️ but code wise  it looks good

PR #963, opened 2020-11-05
Update dd-trace version

johantjellden:
Lets update the datadog package. @palmenhq @hedvigoscar does this look ok?


hedvigoscar:
On condition that you have run it and have not found it to malfunction post dependency update, yes 😉

sampettersson:
I already updated this one a while back, so I think we can close this one :)

PR #1078, opened 2021-02-24
Forward locale to angel data

palmenhq:
What: Allow the option of forwarding the locale string to embark when fetching raw angel data

Why: So we can pick up translations everywhere we fetch angel data


gustaveen (approved):
🙌

PR #1065, opened 2021-02-12
Remove `termsAndConditions` from `Contract` in cross-schema resolver.

niho:
This change is a part of HVG-280 and is dependent on HedvigInsurance/product-pricing#679.

PR #1080, opened 2021-02-25
Revert the schema stitching madness

niho:
Revert "Remove `termsAndConditions` from `Contract` in cross-schema resolver. (#1065)"

This reverts commit 2351a11c0cbee052168173d54ca70156e41b4c32.

## Why

- With the fix in HedvigInsurance/product-pricing#691 we should revert back to using the cross-schema resolution for the old (now deprecated) `termsAndConditions` field on `Contract`. This will keep backwards compatibility.
- This will obviously need to be deployed in synchronicity with the change in product-pricing.

### DIFF
--- a/src/resolvers/cross-schema/index.ts
+++ b/src/resolvers/cross-schema/index.ts
@@ -10,6 +10,7 @@ export const crossSchemaExtensions = `
   extend type Contract {
     perils(locale: Locale!): [PerilV2!]!
     insurableLimits(locale: Locale!): [InsurableLimit!]!
+    termsAndConditions(locale: Locale!): InsuranceTerm!
     insuranceTerms(locale: Locale!): [InsuranceTerm!]!
   }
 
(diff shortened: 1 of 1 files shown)


src/resolvers/cross-schema/index.ts
```diff
@@ -10,6 +10,7 @@ export const crossSchemaExtensions = `
   extend type Contract {
     perils(locale: Locale!): [PerilV2!]!
     insurableLimits(locale: Locale!): [InsurableLimit!]!
+    termsAndConditions(locale: Locale!): InsuranceTerm!
```
FredrikAreschoug, line 13: From my understanding this field is now deprecated let's then be explicit
```suggestion
   @deprecated(reason: "Use `termsAndConditions` form `AgreementCore` instead")
    termsAndConditions(locale: Locale!): InsuranceTerm!
```

sampettersson (approved):
+1 on adding deprecation notice though

hedvigoscar:
Closing this as it has been solved directly in `product-pricing`.

niho:
This is now required again because of HedvigInsurance/product-pricing#707.

PR #1081, opened 2021-02-25
Upgrade apollo server

sampettersson:
Upgrading apollo-server so that we can auto upload the schema to Apollo Studio every time giraffe starts (this is configured by env variables in prod-env later). This will give notifications on when breaking changes are introduced in the schema, etc.

Tested locally by firing some mutations and queries, that worked, will test more thoroughly when this lands in staging.


sampettersson:
`apollo-server-core` didn't compile using `yarn run build` with our old version of TS so upgraded that one also.

FredrikAreschoug (approved):
major release :tada:

PR #1082, opened 2021-02-25
Create a mutation for creating a 'self change'-related quote

bystam:
The difference is that this quote is created after an Embark flow that does not contain all the initial onboarding information, so some of the values are filled by asking member-service for what is missing.

## Why
The upcoming embark flow that produces input for a "self change" (moving flow) quote will not contain all the data that is needed to create a quote. More specifically, it is missing things related to the member herself (name, ssn, email, etc...). In order to plug it into the underwriter quote creation, we must find a way to populate that missing information.

## What
This is a new "self change" related mutation that creates a quote by **only** taking the new quote information related to *what* the insurance contains. The mutation then internally populates the missing values by fetching them from member service, before sending them further to underwriter.

## Screenshot
image

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -63,6 +63,7 @@ interface UserDto {
   selectedCashback: string
   selectedCashbackImageUrl: string
   ssn: string
+  birthDate: string
 }
 
 interface SignDto {
@@ -169,19 +170,20 @@ interface PushTokenDto {
 }
 
 interface CreateQuoteDto {
+  memberId: String
   firstName: string
   lastName: string
-  email?: string
-  ssn?: string
+  ssn: string
+  birthDate: string
   startDate: string
-  incompleteApartmentQuoteData?: {
+  swedishApartmentData?: {
     street: string
     zipCode: string
     householdSize: number
     livingSpace: number
     subType: string
   }
-  incompleteHouseQuoteData?: {
+  swedishHouseData?: {
     street: string
     zipCode: string
     householdSize: number
@@ -196,7 +198,7 @@ interface CreateQuoteDto {
       hasWaterConnected: boolean
     }[]
   }
-  norwegianHomeContents?: {
+  norwegianHomeContentsData?: {
     street: string
     zipCode: string
     coInsured: number
@@ -204,11 +206,11 @@ interface CreateQuoteDto {
     youth: boolean
     subType: string
   }
-  norwegianTravel?: {
+  norwegianTravelData?: {
     coInsured: number
     youth: boolean
   }
-  danishHomeContents?: {
+  danishHomeContentsData?: {
     street: string
     zipCode: string
     livingSpace: number
@@ -216,13 +218,13 @@ interface CreateQuoteDto {
     student: boolean
     subType: string
   }
-  danishAccident?: {
+  danishAccidentData?: {
     street: string
     zipCode: string
     coInsured: number
     student: boolean
   }
-  danishTravel?: {
+  danishTravelData?: {
     street: string
     zipCode: string
     coInsured: number
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -168,28 +168,28 @@ interface PushTokenDto {
   token: string
 }
 
-interface CreateSelfChangeQuoteDto {
+interface CreateQuoteDto {
   firstName: string
   lastName: string
   email?: string
   ssn?: string
   startDate: string
-  swedishApartment?: {
+  incompleteApartmentQuoteData?: {
     street: string
     zipCode: string
     householdSize: number
     livingSpace: number
-    type: string
+    subType: string
   }
-  swedishHouse?: {
+  incompleteHouseQuoteData?: {
     street: string
     zipCode: string
     householdSize: number
     livingSpace: number
-    ancillarySpace: number
+    ancillaryArea: number
     yearOfConstruction: number
     numberOfBathrooms: number
-    isSubleted: boolean
+    subleted: boolean
     extraBuildings: {
       type: string
       area: number
@@ -201,32 +201,32 @@ interface CreateSelfChangeQuoteDto {
     zipCode: string
     coInsured: number
     livingSpace: number
-    isYouth: boolean
-    type: string
+    youth: boolean
+    subType: string
   }
   norwegianTravel?: {
     coInsured: number
-    isYouth: boolean
+    youth: boolean
   }
   danishHomeContents?: {
     street: string
     zipCode: string
     livingSpace: number
     coInsured: number
-    isStudent: boolean
-    type: string
+    student: boolean
+    subType: string
   }
   danishAccident?: {
     street: string
     zipCode: string
     coInsured: number
-    isStudent: boolean
+    student: boolean
   }
   danishTravel?: {
     street: string
     zipCode: string
     coInsured: number
-    isStudent: boolean
+    student: boolean
   }
 }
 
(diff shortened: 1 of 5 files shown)


src/api/index.ts
```diff
@@ -168,6 +168,74 @@ interface PushTokenDto {
...
+    zipCode: string
+    coInsured: number
+    isStudent: boolean
+  }
+  danishTravel?: {
+    street: string
+    zipCode: string
+    coInsured: number
+    isStudent: boolean
+  }
```
bystam, line 230: Looking more closely at the underwriter endpoint, this is of course not at all what that body looks like. It's not far from it, but some adjustments will have to be made to make this work.

src/resolvers/selfChange.ts
```diff
@@ -0,0 +1,70 @@
...
+
+const createSelfChangeQuote: MutationToCreateSelfChangeQuoteResolver = async (
+    _parent,
+    _args,
+    { getToken, headers },
+): Promise<SelfChangeQuoteOutput> => {
+    const input = _args.quoteInput
+    const token = getToken()
+    const member = await getUser(token, headers)
+    console.log(member);
```
palmenhq, line 13: 🙈

src/resolvers/selfChange.ts
```diff
@@ -0,0 +1,70 @@
+
+import { getUser, UserDto, postSelfChangeQuote, CreateQuoteDto } from '../api'
+import { SelfChangeQuoteOutput, MutationToCreateSelfChangeQuoteResolver, SelfChangeQuoteInput } from '../typings/generated-graphql-types'
+
+const createSelfChangeQuote: MutationToCreateSelfChangeQuoteResolver = async (
+    _parent,
+    _args,
+    { getToken, headers },
+): Promise<SelfChangeQuoteOutput> => {
+    const input = _args.quoteInput
```
palmenhq, line 10: if `_args` is not unused i don't think we should prefix it with `_`

palmenhq (commented):
I'm a bit out of the loop here but what's the reasoning for not putting this particular mutation in underwriter directly? Since we've started spreading the mutations and the schema across services it seems weird to me to start implementing stuff in giraffe again

palmenhq:
Awesome work with this though, looks great!

src/resolvers/selfChange.ts
bystam: Oh - I didn't understand that that was a pattern.

bystam:
> I'm a bit out of the loop here but what's the reasoning for not putting this particular mutation in underwriter directly? Since we've started spreading the mutations and the schema across services it seems weird to me to start implementing stuff in giraffe again

Right - so. There are some parts at play here in my mind. I'll try to summarise it in some way that makes sense.

In my mind... I personally actually think that the "all services talk GraphQL and Giraffe stitches it together" is causing a bit of a... brittle design. It makes details of the "upstream" services bleed all the way down to the clients. One big reason it's common to use a "backend for frontend" application is to make meaningful "packaging" of upstream constructs into presentation-layer shapes. This to avoid creating high levels of coupling between back-backend services and consumer clients. So... I personally feel like GraphQL is pretty neat, and makes for good client side consumption of APIs. But I think we lose some strengths of a BFF by actually spearding out the public API in all of our other backend services intead of making it the job of Giraffe.

Secondly - regardless if it's GraphQL or not - I personally also think it furthers the service entanglement if I were to place these additions inside Underwriter. That means drawing even more arrows between upstream services. The big strength of a service like Giraffe is that it can take multiple well defined upstream applications and comine their features like lego bricks into more useful compound features. Both member services and underwriter already have the functionality I need to build the "moving flow" code: 
- member services has the power to give me the "self member" based on session
- and underwriter has a way to create a quote

So I don't actually need to make the "moving flow" feature a construct at all inside any of these two services. Their existing powers contain everything I need. I just need to take two existing, well-defined lego bricks and put them together. I think that's a **perfect** usage of something like Giraffe. But if new features that could be built by just combining existing endpoints were forced to live inside one of the upstream services, then the public API surface of each service would increase rapidly, the amount of arrows pointing between our backend would grow, and the service-to-service couplding would be entrenched.

So... I am aware that there is a large chunk of my opinions going into this implementation. But I at least found a couple of people I discussed this with who felt that I wasn't crazy to try it. 😅

src/resolvers/selfChange.ts
bystam: Eek! Will fix hahaha

src/resolvers/selfChange.ts
palmenhq: It's even a feature of typescript - if you have the `noUnusedParameters` check (which we do) unused paramters are forbidden unless they're prefixed with underscore, indicating that it's unused but required, ie because of the parameter order, such as in this case. One could argue the leading underscore shouldn't be valid for used parameters though... 🙄

palmenhq:
@bystam interesting! 100% see your point and agree here.

I'm just unsure; if this is the direction we're heading - is everyone on board? Or will this become just "another standard"? 😅


**Jokes aside - this is good stuff imo!** And we have similar things in Giraffe already so i don't feel worried this would become an "additional" way of communicating with the backend. But i really think we should think about change management so we don't end up with some people preferring to write the schema in the services and some writing BFF. You get the point i think :) Please challenge this! But i figured it's worth bringing up :)

palmenhq (approved):
👌

bystam:
> @bystam interesting! 100% see your point and agree here.
> 
> I'm just unsure; if this is the direction we're heading - is everyone on board? Or will this become just "another standard"? 😅
> 
> 
> **Jokes aside - this is good stuff imo!** And we have similar things in Giraffe already so i don't feel worried this would become an "additional" way of communicating with the backend. But i really think we should think about change management so we don't end up with some people preferring to write the schema in the services and some writing BFF. You get the point i think :) Please challenge this! But i figured it's worth bringing up :)

I totally agree 100%. I wouldn't have done it 100% rogue, but I have heard from multiple people that they too would like to go in this direction. I feel uncomfortable tagging them here in the PR because maybe I throw them under the bus that way. 😅

Personally I think it's semi-urgent for us to review this approach and align around something. We have had a couple of medium-sized incidents with this only the last few weeks due to it maybe not being suuuuper obvious how schema changes ends up affecting the system.

palmenhq:
Sounds very wise. Maybe this could be a good agenda for tech weekly soon?

hugokallstrom:
Ill chip in from the app team. Agree with you as well Bystam, using giraffe to stitch together different backend services makes sense. Do we still want to talk directly with eg. underwriter for "simpler" requests? Or do we eventually want to use only giraffe as the public API?

bystam:
> Ill chip in from the app team. Agree with you as well Bystam, using giraffe to stitch together different backend services makes sense. Do we still want to talk directly with eg. underwriter for "simpler" requests? Or do we eventually want to use only giraffe as the public API?

I might be wrong here because I haven't read the Android/iOS code - but as I see it, the clients shouldn't necessarily care whether the schema is declared in Giraffe or anywhere else. I would love it if we could consider Giraffe the "only api" that the clients are aware of (at least from a GraphQL standpoint), and whether a piece of the schema comes from Giraffe itself or is attached from an upstream dependency, that's an implementation detail.

Is that not currently how it works?

hugokallstrom:
> > Ill chip in from the app team. Agree with you as well Bystam, using giraffe to stitch together different backend services makes sense. Do we still want to talk directly with eg. underwriter for "simpler" requests? Or do we eventually want to use only giraffe as the public API?
> 
> I might be wrong here because I haven't read the Android/iOS code - but as I see it, the clients shouldn't necessarily care whether the schema is declared in Giraffe or anywhere else. I would love it if we could consider Giraffe the "only api" that the clients are aware of (at least from a GraphQL standpoint), and whether a piece of the schema comes from Giraffe itself or is attached from an upstream dependency, that's an implementation detail.
> 
> Is that not currently how it works?

Yes that is correct, we don't know if the implementation is in giraffe or in an upstream service. It would be nice though, as you say, if giraffe was the only api we needed to care about. Can be a bit confusing atm to keep track of changes in different services, so having giraffe as a kind of middleware would make things easier for us. I think! :)

hugokallstrom (approved):

PR #1096, opened 2021-03-11
Personal number input on danish auth

FredrikAreschoug:
Added personal number as input on `danishBankIdAuth` mutation this is to enable simple sign. 

I know this is breaking the schema but the mutation is not used in prod and for me, it feels better to be explicit that we will need `personalNumber` when authenticating in Denmark. Please let me know if you don't agree and we can discuss it.

### DIFF
--- a/src/features/auth/auth.ts
+++ b/src/features/auth/auth.ts
@@ -45,11 +45,11 @@ const norwegianBankIdAuth: MutationToNorwegianBankIdAuthResolver = async (
 
 const danishBankIdAuth: MutationToDanishBankIdAuthResolver = async (
   _parent,
-  _args,
+  { personalNumber },
   { headers, getToken },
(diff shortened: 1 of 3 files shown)


bystam (approved):
Makes sense to me!

hugokallstrom (approved):

hedvigoscar:
This will break our builds, but we can probably fix it fairly quickly.

Might want to check with iOS before merging though as they are in a bit of a bind wrt builds right now

tarikstafford:
> This will break our builds, but we can probably fix it fairly quickly.
> 
> Might want to check with iOS before merging though as they are in a bit of a bind wrt builds right now

I think it's safe but not 100% sure. We should wait for @sampettersson who is back tomorrow.

PR #1112, opened 2021-03-26
move SelfChangeEligibility to giraffe to place it more symmetrically with the mutation

bystam:
JIRA ticket: [BFF-22]

[BFF-22]: https://hedvig.atlassian.net/browse/BFF-22

### DIFF
--- a/src/schema.graphqls
+++ b/src/schema.graphqls
@@ -13,6 +13,8 @@ type Query {
   chatActions: [ChatAction]
   geo: Geo!
   angelStory(name: String!, locale: String): AngelStory
+  """Returns a type describing whether the 'Self Change' functionality is possible."""
+  selfChangeEligibility: SelfChangeEligibility!
 }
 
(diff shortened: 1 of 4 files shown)


src/schema.graphqls
```diff
@@ -325,6 +327,34 @@ input CampaignInput {
   name: String
 }
 
+type SelfChangeEligibility {
+    """A list of reasons for why 'Self Change' is not possible - if empty 'Self Change' can be done."""
+    blockers: [SelfChangeBlocker!]! @deprecated(reason: "Use embarkStoryId to determine if it is eligible or not")
```
hedvigoscar, line 332: With a little luck, we can actually hard-remove this - it only needs to be present in the schema when we build the apps, consumers can still function without it being in the schema (as consumers never fetch the schema run-time). I'll make sure to investigate that!

hedvigoscar (approved):
lgtm

PR #1114, opened 2021-03-29
perform embark story triaging based on the market of the member

bystam:
# Jira ticket [BFF-22]

### What
- Fill out the real values of the `embarkStoryId` based on the market of the member

### Why
- This should not be hard-coded in the clients, but since it's very UI-centric, it makes sense to live here.

[BFF-22]: https://hedvig.atlassian.net/browse/BFF-22

### DIFF
--- a/src/resolvers/selfChange.ts
+++ b/src/resolvers/selfChange.ts
@@ -8,7 +8,7 @@ import {
     SelfChangeEligibility
 } from '../typings/generated-graphql-types'
 
-const storiesByMarket: { [key: string]: string } = {
+const storiesByMarket: Record<string, string> = {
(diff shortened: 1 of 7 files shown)


src/resolvers/selfChange.ts
```diff
@@ -8,14 +8,26 @@ import {
     SelfChangeEligibility
 } from '../typings/generated-graphql-types'
 
+const storiesByMarket: { [key: string]: string } = {
```
palmenhq, line 11: a neater type for this imo would be `Record<string, string>`, it means the same thing

palmenhq (approved):
Looks great! really good job with this!

PR #1115, opened 2021-03-29
Build a little more streamlined alternative to the existing 'callApi' upstream functions

bystam:
## Background
Calling simple upstream endpoints in Giraffe is somewhat cumbersome. When writing a resolver function for GraphQL, we have to dig out somewhat low level constructs like `getToken` and `headers`, which we forward to every upstream call. We never need to specifically work with these in any particular way, we just need to forward them.

Similarly, when declaring new upstream calls, the same object is constructed and copy pasted between each single call:
```typescript
{
    mergeOptions: {
      headers: (headers as any) as RequestInit['headers'],
      method: 'POST',
      body: JSON.stringify(body),
    },
    token
  }
```

There was also no way to actually point the `callApi` implementation towards the local environment, since it depended on a `BASE_URL` approach, which wasn't properly compatible with how we use ports to discern services locally, but path prefixes when doing so remote.

Lastly, having all the API calls as top level functions from the same file generates quite a lot of imports for each method.

Wouldn't it be nicer if doing upstream calls was less cumbersome and reminded you of working with `FeignClients`? :) 

## Solution
This suggestion basically is built on top of having `HttpClient` types that has `callApi`:esque functions that **already has the token and headers pre-packaged**. On top of that, there are types leveraging that client to build nice top level functions for specific API calls. These are sorted into which upstream service they target, and are grouped in different files.

This injected "umbrella type" lives inside the context each resolver receives, and contains a `ProductPricingClient` for now:
```typescript
export interface Upstream {
    productPricing: ProductPricingClient
    // TODO: add underwriter, memberService and friends
}
```

```typescript
// no token/headers needed, that's baked into the implementation of this type
export interface ProductPricingClient {
  getContractMarketInfo(): Promise<ContractMarketInfoDto>
  getSelfChangeEligibility(): Promise<SelfChangeEligibilityDto>
}
```

And here's what it looks like at the call site:
```typescript
const myGraphQLQuery: QueryToSomething = async (
    _parent,
    _args,
    { upstream }, // look!!! an `Upstream` 
): Promise<ResultThingy> => {
  // neat!
  // and the best part is I don't need to spell out any imports to get this beyond the auto-generated
  // one from the GraphQL resolver signature 
  const info = await upstream.productPricing.getContractMarketInfo()
}
```

### Adding new upstream calls
Based on this, it's super easy to add new upstream calls, especially if there's already a client for the service you're targeting:

Let's say we want to add the ability to GET and POST animals to an animal API. Here we go:
```typescript
// declare some data types:
interface AnimalClient {
  get(id: string): Promise<Animal>
  create(animal: Animal): Promise<Animal>
}

interface Animal {
  id?: string
  name: string
}

// and later in the implementation
{ // client here is the injected HttpClient
  get: async (id) => {
    const res = await client.get(`/animals/${id}`)
    return res.json()
  },
  create: async (animal) => {
    const res = await client.post("/animals", animal)
    return res.json()
  }
}

```

### Bonus features
- This thing can call your local machine, `callApi` couldn't due to its baseUrl setup
- Whether you want to target the local upstream or staging can be toggled using `UPSTREAM_MODE="local|remote"` in `.env`
- When targeting your local environment, you can pretend to be a specific user by setting `LOCAL_MEMBERID_OVERRIDE=<member-id>` in `.env`

### DIFF
--- a/src/api/httpClient.ts
+++ b/src/api/httpClient.ts
@@ -32,7 +32,7 @@ export const createContextfulHttpClient = (
 
     const requestOptions: RequestInit = {
       method: method,
-      body: body !== undefined ? JSON.stringify(body) : body,
+      body: body ? JSON.stringify(body) : body,
       headers: headers,
     }
 
--- a/src/api/httpClient.ts
+++ b/src/api/httpClient.ts
@@ -1,5 +1,6 @@
 
 import fetch, { RequestInit, Response } from 'node-fetch'
+import { TokenProvider } from '../context'
 
 /**
  * A client that can perform upstream HTTP calls. It can be assumed
@@ -18,7 +19,7 @@ export interface HttpClient {
  */
 export const createContextfulHttpClient = (
   baseUrl: string,
-  getToken: () => string,
+  getToken: TokenProvider,
   forwardHeaders: { [key: string]: string }
 ): HttpClient => {
   const call = async (url: string, method: string, body: any | undefined = undefined): Promise<Response> => {
--- a/src/api/httpClient.ts
+++ b/src/api/httpClient.ts
@@ -13,6 +13,13 @@ export interface HttpClient {
   delete(url: string): Promise<Response>
 }
 
+export class HttpError extends Error {
+  constructor(public statusCode: number, public body: any) {
+    super(`HTTP error - statusCode ${statusCode}`)
+    this.name = "HttpError"
+  }
+}
+
 /**
  * Create a HttpClient that comes with a pre-packaged base-url,
  * auth token and headers.
@@ -22,7 +29,7 @@ export const createContextfulHttpClient = (
   getToken: TokenProvider,
   forwardHeaders: { [key: string]: string }
 ): HttpClient => {
-  const call = async (url: string, method: string, body: any | undefined = undefined): Promise<Response> => {
+  const call = (method: RequestInit['method']) => async (url: string, body?: unknown): Promise<Response> => {
     const headers: { [key: string]: string } = {
       ...forwardHeaders,
       Accept: 'application/json',
@@ -40,19 +47,17 @@ export const createContextfulHttpClient = (
 
     const res = await fetch(`${baseUrl}${url}`, requestOptions)
 
-    if (res.status >= 400) { // TODO: include custom status validation when needed
-      throw new Error(
-        `API error - status: ${res.status}, body: ${JSON.stringify(await res.text(), null, 4)}`
-      )
+    if (res.status >= 400) {
+      throw new HttpError(res.status, await res.json())
     }
 
     return res
   }
 
   return {
-    get: (url) => call(url, "GET"),
-    post: (url, body) => call(url, "POST", body),
-    put: (url, body) => call(url, "PUT", body),
-    delete: (url) => call(url, "DELETE"),
+    get: call('GET'),
+    post: call('POST'),
+    put: call('PUT'),
+    delete: call('DELETE'),
   }
 }
--- a/src/api/httpClient.ts
+++ b/src/api/httpClient.ts
@@ -17,14 +17,14 @@ export interface HttpClient {
  */
 export const createContextfulHttpClient = (
   baseUrl: string,
-  token: string,
+  getToken: () => string,
   forwardHeaders: { [key: string]: string }
 ): HttpClient => {
   const call = async (url: string, method: string, body: any | undefined = undefined): Promise<Response> => {
     const headers: { [key: string]: string } = {
       ...forwardHeaders,
       Accept: 'application/json',
-      Authorization: `Bearer ${token}`
+      Authorization: `Bearer ${getToken()}`
     }
     if (body) {
       headers['Content-Type'] = 'application/json'
--- a/src/api/httpClient.ts
+++ b/src/api/httpClient.ts
@@ -1,0 +1,51 @@
+
+import fetch, { RequestInit, Response } from 'node-fetch'
+
+/**
+ * A client that can perform upstream HTTP calls.
+ */
+export interface HttpClient {
+  get(url: string): Promise<Response>
+  post(url: string, body: any): Promise<Response>
+  put(url: string, body: any): Promise<Response>
+  delete(url: string): Promise<Response>
+}
+
+/**
+ * Create a HttpClient that comes with a pre-packaged base-url,
+ * auth token and headers.
+ */
+export const createContextfulHttpClient = (
+  baseUrl: string,
+  token: string,
+  forwardHeaders: { [key: string]: string }
+): HttpClient => {
+  const call = async (url: string, method: string, body: any | undefined = undefined): Promise<Response> => {
+    const headers: { [key: string]: string } = {
+      ...forwardHeaders,
+      Accept: 'application/json',
+    }
+    if (body) {
+      headers['Content-Type'] = 'application/json'
+    }
+    if (token) {
+      headers.Authorization = `Bearer ${token}`
+    }
+
+    const requestOptions: RequestInit = {
+      method: method,
+      body: body !== undefined ? JSON.stringify(body) : body,
+      headers: headers,
+    }
+
+    const res = await fetch(`${baseUrl}${url}`, requestOptions)
+    return res
+  }
+
+  return {
+    get: (url) => call(url, "GET"),
+    post: (url, body) => call(url, "POST", body),
+    put: (url, body) => call(url, "PUT", body),
+    delete: (url) => call(url, "DELETE"),
+  }
+}
--- a/src/api/httpClient.ts
+++ b/src/api/httpClient.ts
@@ -32,7 +32,7 @@ export const createContextfulHttpClient = (
 
     const requestOptions: RequestInit = {
       method: method,
-      body: body ? JSON.stringify(body) : body,
+      body: body ? JSON.stringify(body) : undefined,
       headers: headers,
     }
 
--- a/src/api/httpClient.ts
+++ b/src/api/httpClient.ts
@@ -2,7 +2,8 @@
 import fetch, { RequestInit, Response } from 'node-fetch'
 
 /**
- * A client that can perform upstream HTTP calls.
+ * A client that can perform upstream HTTP calls. It can be assumed
+ * to contain the necessary headers to fire calls in the current context.
  */
 export interface HttpClient {
   get(url: string): Promise<Response>
--- a/src/api/httpClient.ts
+++ b/src/api/httpClient.ts
@@ -24,12 +24,10 @@ export const createContextfulHttpClient = (
     const headers: { [key: string]: string } = {
       ...forwardHeaders,
       Accept: 'application/json',
+      Authorization: `Bearer ${token}`
     }
     if (body) {
       headers['Content-Type'] = 'application/json'
-    }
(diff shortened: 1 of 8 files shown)


src/context.ts
```diff
@@ -33,18 +35,20 @@ const getWebContext = (graphCMSSchema: GraphQLSchema) => async ({
...
     getToken,
     graphCMSSchema,
-    headers: {
-      'User-Agent': checkedCtx.get('User-Agent'),
-      'X-Forwarded-For': checkedCtx.get('x-forwarded-for'),
-      'X-Request-Id': checkedCtx.get('x-request-id') || uuidv4(),
-      'Accept-Language': checkedCtx.get('accept-language'),
-      'Enable-Simple-Sign': checkedCtx.get('Enable-Simple-Sign'),
-    },
+    headers: headers,
```
vonElfvin, line 48:
```suggestion
    headers,
```

src/api/upstreamApi.ts
```diff
@@ -0,0 +1,44 @@
...
+            if (config.LOCAL_MEMBERID_OVERRIDE) {
+                allHeaders["Hedvig.token"] = config.LOCAL_MEMBERID_OVERRIDE as string
+            }
+        }
+        return createContextfulHttpClient(baseUrl, token, allHeaders)
+    }
+
+    return {
+        productPricing: createProductPricingClient(
+            createHttpClient("/productPricing", 4085)
```
vonElfvin, line 41: Is hard coding 4085 risky?
```suggestion
            createHttpClient("/productPricing", 4085)
```

src/api/upstreamApi.ts
```diff
@@ -0,0 +1,44 @@
...
+    const createHttpClient = (remotePathPrefix: string, localPort: number): HttpClient => {
+        const allHeaders = (headers as any) as { [key: string]: string }
+
+        let baseUrl: string = `${config.BASE_URL}${remotePathPrefix}`
+        if (config.UPSTREAM_MODE == "local") {
+            baseUrl = `http://localhost:${localPort}`
+            if (config.LOCAL_MEMBERID_OVERRIDE) {
+                allHeaders["Hedvig.token"] = config.LOCAL_MEMBERID_OVERRIDE as string
+            }
+        }
```
vonElfvin, line 35: `let` 🔪 
```suggestion
        const baseUrl: string = config.UPSTREAM_MODE == "local" ? `http://localhost:${localPort}` : `${config.BASE_URL}${remotePathPrefix}`
        if ( config.UPSTREAM_MODE == "local" && config.LOCAL_MEMBERID_OVERRIDE) {
            allHeaders["Hedvig.token"] = config.LOCAL_MEMBERID_OVERRIDE as string
        }
```

src/api/upstreamApi.ts
```diff
@@ -0,0 +1,44 @@
...
+): UpstreamApi => {
+
+    const createHttpClient = (remotePathPrefix: string, localPort: number): HttpClient => {
+        const allHeaders = (headers as any) as { [key: string]: string }
+
+        let baseUrl: string = `${config.BASE_URL}${remotePathPrefix}`
+        if (config.UPSTREAM_MODE == "local") {
+            baseUrl = `http://localhost:${localPort}`
+            if (config.LOCAL_MEMBERID_OVERRIDE) {
+                allHeaders["Hedvig.token"] = config.LOCAL_MEMBERID_OVERRIDE as string
```
vonElfvin, line 33: I have no idea if this is the case, just to make sure
```suggestion
                allHeaders["Hedvig.Token"] = config.LOCAL_MEMBERID_OVERRIDE as string
```

src/api/httpClient.ts
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Create a HttpClient that comes with a pre-packaged base-url,
+ * auth token and headers.
+ */
+export const createContextfulHttpClient = (
+  baseUrl: string,
+  token: string,
+  forwardHeaders: { [key: string]: string }
+): HttpClient => {
+  const call = async (url: string, method: string, body: any | undefined = undefined): Promise<Response> => {
```
vonElfvin, line 23: If you want
```suggestion
  const call = async (url: string, method: string, body?: any = undefined): Promise<Response> => {
```

src/api/httpClient.ts
```diff
@@ -0,0 +1,51 @@
...
+    const headers: { [key: string]: string } = {
+      ...forwardHeaders,
+      Accept: 'application/json',
+    }
+    if (body) {
+      headers['Content-Type'] = 'application/json'
+    }
+    if (token) {
```
vonElfvin, line 31: Token is not "undefined-able" in the function, maybe add a `?` or `string | null` depending on what makes sense

src/api/upstreams/productPricing.ts
```diff
@@ -0,0 +1,29 @@
...
+export const createProductPricingClient = (client: HttpClient): ProductPricingClient => {
+  return {
+    getContractMarketInfo: async () => {
+      const res = await client.get("/contracts/market-info")
+      return res.json()
+    },
+    getSelfChangeEligibility: async () => {
+      const res = await client.get("/contracts/selfChange/eligibility")
+      return res.json()
+    }
```
vonElfvin, line 27: Do we use this for anything?

src/context.ts
```diff
@@ -77,6 +81,7 @@ const getWebSocketContext = (graphCMSSchema: GraphQLSchema) => (
     remoteIp:
       headers['X-Forwarded-For'] ||
       (context.request.connection.address() as string),
+    upstream: createUpstreamApi(getToken(), headers)
```
vonElfvin, line 84: Not sure exactly how this is used, but I see a risk in `upstream` being set at startup and thus `getToken()` being called only once, thus not being refreshed based on whoever makes the call. This is probably not the case though 🤷

vonElfvin (commented):
This looks neat and useful, added some code stuff I thought of while reading through

src/api/upstreamApi.ts
bystam: I personally wouldn't think so? This is only the port used when targeting localhost. Do people usually change the ports of their services locally? In that case, I think we could extend it later to accomodate for that.

If we turn this into an env var or something I think it could give the false indication that it varies, when it is actually quite constant.

src/api/upstreamApi.ts
bystam: [HTTP headers are case insensitive](https://stackoverflow.com/a/5259004) - so it shouldn't matter :) I just chose the casing I saw most common in the backend for consistency.

src/api/httpClient.ts
bystam: Tried it - doesn't work actually
image

src/api/upstreams/productPricing.ts
bystam: Yes - it's used inside the self change queries to determine whether the "embark story" should be returned or not. This was reorganised a bit last week to make it a bit more clear which service was in charge of what, and to avoid a really obscure cross schema stiching thing me and Oscar looked into.

src/context.ts
bystam: No - based on my understanding (and testing) this is evaluated once upon each GraphQL "operation". I.e. each HTTP call. Therefore it will contain the correct values for each user and request.

src/context.ts
```diff
@@ -77,6 +81,7 @@ const getWebSocketContext = (graphCMSSchema: GraphQLSchema) => (
     remoteIp:
       headers['X-Forwarded-For'] ||
       (context.request.connection.address() as string),
+    upstream: createUpstream(getToken(), headers)
```
bystam, line 84: I don't really understand why `getToken()` is a function we inject, since it's synchronous and always re-created on every request that hits Giraffe. Feels like it could just as easily just be a string.

bystam: Oh wait - it's because some upstream requests don't need it maybe? I guess I should inject this lazily into the `HttpClient` too so that it doesn't break existing things that don't need authorization?

palmenhq (commented):
Awesome stuff and well-needed improvements! I have some questions though :)

src/api/upstream.ts
```diff
@@ -0,0 +1,43 @@
...
+): Upstream => {
+
+    const createHttpClient = (remotePathPrefix: string, localPort: number): HttpClient => {
+        const allHeaders = (headers as any) as { [key: string]: string }
+
+        const baseUrl: string = config.UPSTREAM_MODE == "local"
+            ? `http://localhost:${localPort}`
+            : `${config.BASE_URL}${remotePathPrefix}`
+        if (config.UPSTREAM_MODE == "local" && config.LOCAL_MEMBERID_OVERRIDE) {
+            allHeaders["Hedvig.token"] = config.LOCAL_MEMBERID_OVERRIDE as string
```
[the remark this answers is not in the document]
bystam: I wrote an answer to Elvin about that. Headers are case insensitive in the HTTP spec. And we use different casing in several places in our systems :)

src/api/httpClient.ts
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Create a HttpClient that comes with a pre-packaged base-url,
+ * auth token and headers.
+ */
+export const createContextfulHttpClient = (
+  baseUrl: string,
+  getToken: () => string,
+  forwardHeaders: { [key: string]: string }
+): HttpClient => {
+  const call = async (url: string, method: string, body: any | undefined = undefined): Promise<Response> => {
```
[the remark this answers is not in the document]
bystam: Haha yes - I thought about that, I just didn't feel functionally cool enough. But I can try it :)

src/api/httpClient.ts
```diff
@@ -0,0 +1,56 @@
+
+import fetch, { RequestInit, Response } from 'node-fetch'
+
+/**
+ * A client that can perform upstream HTTP calls.
+ */
```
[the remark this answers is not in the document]
bystam: It doesn't contain "upstream", but yes, I guess :)

src/api/upstream.ts
```diff
@@ -0,0 +1,43 @@
...
+ * ...
+ * const quote = await upstream.underwriter.getQuote("id")
+ * ```
+ */
+export interface Upstream {
+    productPricing: ProductPricingClient
+}
+
+export const createUpstream = (
+    token: () => string,
```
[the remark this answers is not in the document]
bystam: Maybe? Like a functional type or something? Because it's currently injected like this and used like `getToken()` in dozens of places.

src/api/upstream.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    const createHttpClient = (remotePathPrefix: string, localPort: number): HttpClient => {
+        const allHeaders = (headers as any) as { [key: string]: string }
+
+        const baseUrl: string = config.UPSTREAM_MODE == "local"
+            ? `http://localhost:${localPort}`
+            : `${config.BASE_URL}${remotePathPrefix}`
+        if (config.UPSTREAM_MODE == "local" && config.LOCAL_MEMBERID_OVERRIDE) {
+            allHeaders["Hedvig.token"] = config.LOCAL_MEMBERID_OVERRIDE as string
+        }
```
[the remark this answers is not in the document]
bystam: I would love to work with you IRL to improve upon this. I will bring it up on standup.

PR #1140, opened 2021-04-15
Proxy embark's format.js file

palmenhq:
so we can go to `giraffe.hedvig.com/embark/format.js` instead of the heroku app


src/index.ts
```diff
@@ -90,11 +89,11 @@ makeSchema()
...
       app.use(
-        route.get(
-          '/embark/format.js',
-          // @ts-ignore - False positive
-          proxy(process.env.EMBARK_FORMAT_ENDPOINT, {}),
-        ),
+        proxy('/embark/format.js', {
+          target: process.env.EMBARK_FORMAT_ENDPOINT,
+          logs: true,
+          rewrite: (_path) => '', // the path should always be exactly '/embark/format.js'
```
styrelius, line 95: I don't really understand what this is and why it's needed 🙃

styrelius (commented):
I don't really have any idea how to test this locally and I know nothing about how giraffe works so I don't really think I should be the one approving this 😔

src/index.ts
palmenhq: It lets you modify the proxied path :) we're proxying from `/embark/format.js` (line 92) to `http://embark/format.js` (from `process.env.EMBARK_FORMAT_ENDPOINT`, see .env.example). Normally you'd want to forward the path to your proxy endpoint, but as we're specifying `/format.js` in there already we don't want the path to follow through, because then we'd try to reach `http://embark/format.js/embark/format.js` (not correct). `rewrite` returning an empty string here will remove the `/embark/format.js` from the path, meaning the absolute url we forward the request to becomes `http://embark/format.js` (correct).

PR #1141, opened 2021-04-21
Quote details table

sampettersson:
This PR extends giraffe with a cross schema extension that on a quote builds a table with labels and values that allows the app to easily render it without digging through quote details, this will be used to drive this component in the apps:

Screenshot 2021-04-21 at 11 27 04


src/translations/da_DK.json
```diff
@@ -0,0 +1,8 @@
+{
```
hedvigoscar, line 1: Should we gitignore these and download them in CI? 🤔 single source of truth and all

hedvigoscar (approved):
I like it!

src/translations/da_DK.json
sampettersson: @hedvigoscar web-onboarding did it like this, not really sure how we should approach it 🤔 it's kinda nice to have them checked in to not have to fiddle too much imo

But we can change it later ofc.

hugokallstrom (commented):
Looks good! Could we take 5-10 minutes after standup tomorrow and go through this? Not 100% on how this is connected to other services.

sampettersson:
@hugokallstrom absolutely :)

PR #1148, opened 2021-04-23
MX-94: A simpler "moving flow" mutation in Giraffe, offloading client side logic

bystam:
### What
- Replace the old `createSelfChangeQuote` mutation with a much simpler `createAddressChangeQuotes` that focuses on what the client wants:
  - Posting a new address
  - Receiving a list of quotes back
- Makes giraffe in charge of turning the high level concept of "changing address" to the lower level concepts of "creating one or more quotes based on the address change"

### Why
- Wiring the existing mutation to Embark apparently was a bit of a challenge, and I then realised that we can make it much simpler by letting Giraffe be the middle-man between low level and high level concepts.


package.json
```diff
@@ -59,8 +61,11 @@
     "apollo": "^2.16.3",
     "concurrently": "^4.1.1",
     "graphql-schema-typescript": "^1.2.9",
+    "jest": "^24.9.0",
+    "jest-expo": "^41.0.0",
     "nodemon": "^1.19.1",
     "prettier": "^1.18.2",
+    "ts-jest": "^24.2.0",
```
bystam, line 68: Jest versions directly copied from embark 🤷

src/schema.graphqls
```diff
@@ -328,9 +328,10 @@ input CampaignInput {
 }
 
 type SelfChangeEligibility {
-    """A list of reasons for why 'Self Change' is not possible - if empty 'Self Change' can be done."""
-    blockers: [SelfChangeBlocker!]! @deprecated(reason: "Use embarkStoryId to determine if it is eligible or not")
-    embarkStoryId: ID
+    blockers: [SelfChangeBlocker!]! @deprecated(reason: "Use addressChangeEmbarkStoryId instead")
+    embarkStoryId: ID @deprecated(reason: "Use addressChangeEmbarkStoryId instead")
+    """The ID of an embark story that contains an address change flow, if eligible."""
+    addressChangeEmbarkStoryId: ID
```
bystam, line 334: `SelfChangeEligibility` remains the umbrella typ explaining self change eligibility, and then it contains "routing IDs" for the address change flow. This model made sense to me at least.

src/resolvers/selfChange.ts
```diff
@@ -1,98 +1,61 @@
...
-            ...input.danishAccident
-        }: undefined
-    }
+const createAddressChangeQuotes: MutationToCreateAddressChangeQuotesResolver = async (
+  _parent,
+  args,
+  { upstream },
+): Promise<AddressChangeOutput> => {
+  const member = await upstream.memberService.getSelfMember()
+  const contracts = await upstream.productPricing.getMemberContracts()
```
bystam, line 46: Let me know what you think, I personally like reading the upstream HTTP calls this way :)

src/api/upstreams/productPricing.ts
```diff
@@ -2,21 +2,42 @@
...
+  TERMINATED_IN_FUTURE = 'TERMINATED_IN_FUTURE',
+  TERMINATED = 'TERMINATED'
+}
+
 export const createProductPricingClient = (client: HttpClient): ProductPricingClient => {
   return {
+    getMemberContracts: async () => {
+      const res = await client.get("/contracts")
+      return res.json()
+    },
```
bystam, line 40: This endpoint doesn't exist yet, but I intend to build it.

src/schema.graphqls
```diff
@@ -355,118 +356,46 @@ enum SelfChangeBlocker {
...
-    OUTHOUSE
-    GUESTHOUSE
-    GAZEBO
-    GREENHOUSE
-    SAUNA
-    BARN
-    BOATHOUSE
-    OTHER
-}
+  countryCode: String!
```
hedvigoscar, line 367: Should this be a `Market` of some sort? 🤔

src/schema.graphqls
```diff
@@ -355,118 +356,46 @@ enum SelfChangeBlocker {
...
+    type: String!
+    area: Int!
+    hasWaterConnected: Boolean!
 }
 
-type SelfChangeQuoteOutput {
-  id: ID!
-  price: Int!
-  validTo: TimeStamp!
+type AddressChangeOutput {
```
hedvigoscar, line 397: How do we communicate if the new quote cannot be created due to hitting underwriting limits, or the information provided is insufficient?

src/schema.graphqls
```diff
@@ -355,118 +356,46 @@ enum SelfChangeBlocker {
     TOO_MANY_CONTRACTS
 }
 
-input SelfChangeQuoteInput {
+input AddressChangeInput {
```
hedvigoscar, line 359: How can I as a consumer of the API know what information I need to provide to create an Address Change-quote? Do we have a path for gracefully changing what data points need to be gathered?

hedvigoscar (commented):
I added some concerns 🙂

src/schema.graphqls
bystam: I don't know - maybe? Usually I find enumifying country codes to be a bit of an anti-pattern, where one introduces new spelling for an already established and stable concept. But that mostly a problem to me if one also lets that value spill into other systems, where we end up having to change multiple codebases to add new enum cases.

I guess we could call this thing `market` and let it be spelled as an ISO-country code though.

src/schema.graphqls
bystam: > How do we communicate if the new quote cannot be created due to hitting underwriting limits

I noticed that this was a thing just now. The fact that this happens is buried within a `: Any` return type in the backend. I guess the return value will have to become some kind of union type here?

And since we create more than one quote I guess we'll have to make some kind of union of their guideline breaching problems?

src/schema.graphqls
bystam: Doesn't GraphQL come with some of the out of the box? Or do you mean dynamically somehow pre-emptively explaining what data is needed through the API itself? That sounds a lot like rewriting our APIs in some kind of real restful HATEOS architecture. But that would also mean having to make the Embark stories dynamic based on it? Or?

src/resolvers/selfChange.ts
hedvigoscar: Since these do not have to be run in sequence, it's possible to use the following form to parallelize the work:

```ts
const [
  member,
  contracts,
] = await Promise.all[
  upstream.memberService.getSelfMember(),
  upstream.productPricing.getMemberContracts(),
]
```

src/schema.graphqls
hedvigoscar: I'm a big fan of types constraining their values to what's actually valid, and I think there's a very small subset of `String` that is actually valid here 😧

src/schema.graphqls
hedvigoscar: >I noticed that this was a thing just now. The fact that this happens is buried within a : Any return type in the backend.

Smells like the underpowered Java type-system 😞

>I guess the return value will have to become some kind of union type here?

That would be nice, I think!

>And since we create more than one quote I guess we'll have to make some kind of union of their guideline breaching problems?

I'm guessing that the entire transaction fails if any of the quotes hit underwriting limits, which is really what we're interested in here as the consumers of the API (did it work or did it not, basically).

Hence, we could perhaps go with something like this:


```graphqls
union AddressChange{Outcome|Result|WhateverIsYourPreferredTerminology} = 
    AddressChangeSuccess | AddressChangeUnderwritingLimitsHit

type AddressChangeSuccess {
    quoteIds: [ID!]!
}

type AddressChangeUnderwritingLimitsHit {
    reasons: [String!]! # This is probably optional to implement given that there's no use-case for it
}
```

src/schema.graphqls
hedvigoscar: There's been long discussions about `input union` in the GraphQL spec, but it does not exist.

Given the constraints we're working with in calling GraphQL mutations from Embark they may not make sense, either.

Perhaps we cannot make it type-level documentation, but I think it would be valid to have it documented _somewhere_, such that do not let it be knowledge in the head of the implementing developers, but instead exist as knowledge in the world.

It would be slightly thankless to come across this mutation and to have to trial-and-error your way into finding the right subset of fields that you need or do not need to provide in order to successfully call this mutation 😕

src/schema.graphqls
bystam: > I'm guessing that the entire transaction fails if any of the quotes hit underwriting limits

No, that's not the case I'm afraid. Quotes are created individually, so we won't be able to run all of them in any kind of transaction.

hedvigoscar: Sorry - what I meant was that the 'transaction' we're creating here, ie. creating a set of quotes to be signed upon moving, is to be considered failed if any of the quotes we attempt to create fails - we can't really offer a partial move - right?

src/schema.graphqls
bystam: I agree with that in general, but if that means turning country codes into an enum I disagree, unless we explicitly add all the cases in the world to that enum upfront. I think using enums for things that expand frequently is a bit of a code smell, and is not really what enums are designed for in the first place.

I think the name itself explains quite clearly that it will only work if it's an actual country code. I guess I could prefix it with `iso` to be even more clear.

----

Then again, maybe my opinion here more applies to backend systems where the values are passed around many API barriers, where introducing enums creates a lot of brittleness in the system. Then, introducing a case in one place starts breaking other places have have yet to be updated. Simply using either strings, or string wrapping classes, makes these things way more resilient in those contexts. Also, again, I feel that enums usually should be limited to types that have a limited and frozen set of enumerable cases.

Buuuut - since this is only used immediately inside the resolver, and does not leak to any upstream service, I could make an exception to my conviction and turn it into `Market.SE` and `Market.NO` in this case :)

src/schema.graphqls
bystam: But we can still consider the whole thing as something that fails if a single one of them does. Creating quotes is a lightweight thing that does not mutate anything in the system really, anyway.

src/schema.graphqls
bystam: Yes, right. I think documentation is perfect for that right now. Right now I was mostly aiming for this to be something that we wire to the embark flows, which will be trivial. Due to the somewhat hacky nature of how moving flow is implemented in the backend, I doubt this will be a reusable thing that we build empires upon.

src/schema.graphqls
hedvigoscar: GraphQL is explicitly designed to handle API evolution with Enums - all the code generation-tooling will generate an `UNKNOWN`-value to handle Enum-values that were not present at the time of fetching the schema 🙂

src/schema.graphqls
hedvigoscar: I think we mean the same thing - I think I used the word transaction poorly as the word implies rollback. I meant that we need to fail the entire request if any of the created quotes hit UW-limits, and I think you mean that as well 🙂

src/schema.graphqls
hedvigoscar: In that case, let's add some inline documentation (comment before the type definition should do the trick), and keep the horizon for sunsetting the code in mind at all times 🙂

src/schema.graphqls
bystam: Cool!

src/schema.graphqls
bystam: Yes! Agreed :)

src/schema.graphqls
bystam: I added it now! :)

src/api/httpClient.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    const result = client.get('/fail-test')
+
+    expect(result).rejects.toThrowError()
+  })
+
+  it('Does not fail if status code is overridden', async () => {
+    const result = await client.get('/fail-test', { validStatusCodes: [400] })
+
+    expect(await result.json()).toEqual({ result: 'FAIL' })
+  })
```
bystam, line 81: A bit bare tests right now, but actually setting up a webserver and running towards it makes for a pretty nice way to verify that the `HttpClient` thing actually works :)

src/resolvers/selfChange.ts
```diff
@@ -1,98 +1,97 @@
...
+  { upstream },
+): Promise<AddressChangeQuoteResult> => {
+  const contractIds = args.input.contractBundleId.replace('bundle:', '').split(',')
+  const [
+    member,
+    contracts,
+    marketInfo
+  ] = await Promise.all([
+    upstream.memberService.getSelfMember(),
+    Promise.all(contractIds.map(upstream.productPricing.getContract)),
```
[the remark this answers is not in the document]
bystam: Won't that make the destructuring array different though? I want the array I'm destructuring into be 3 elements, where the middle element is an array of contracts.

src/schema.graphqls
```diff
@@ -53,7 +53,8 @@ type Mutation {
   registerBranchCampaign(campaign: CampaignInput!): Boolean
   updateLanguage(input: String!): Boolean! # input should be in the format as Accept-Language header
   updatePickedLocale(pickedLocale: Locale!): Member!
-  createSelfChangeQuote(quoteInput: SelfChangeQuoteInput!): SelfChangeQuoteOutput!
+  """Create all the quotes needed in relation to a change of address, based on the current state of the member's insurance."""
```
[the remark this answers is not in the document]
bystam: Wait what. I thought that triple-quotes were documentation and `#` were actual comments. But you're saying that  `#` is actual documentation comments too?

src/resolvers/selfChange.ts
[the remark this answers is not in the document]
hedvigoscar: Heh. Yes. My bad 🤦

PR #1152, opened 2021-04-27
Upcoming and current contract agreement extension

tarikstafford:


src/resolvers/cross-schema/quoteDetailsTable.ts
```diff
@@ -42,61 +43,9 @@ export const crossSchemaExtensions = `
...
-            }
-            ... on DanishAccidentDetails {
-                coInsured
-            }
-            ... on DanishTravelDetails {
-                coInsured
-            }
-          }
-        }`,
+        fragment: fragment,
```
sampettersson, line 48:
```suggestion
        fragment,
```

src/resolvers/cross-schema/index.ts
```diff
@@ -488,8 +488,8 @@ export const getCrossSchemaResolvers = (
           })
         },
       },
-      ...quoteDetailsTable("BundledQuote"),
-      ...quoteDisplayName("BundledQuote")
+      ...quoteDetailsTable(getQuoteDetailsFragment("BundledQuote")),
+      ...quoteDisplayName(getQuoteDetailsFragment("BundledQuote"))
```
sampettersson, line 492:
```suggestion
      ...quoteDisplayName("BundledQuote")
```

src/resolvers/cross-schema/index.ts
```diff
@@ -405,8 +405,8 @@ export const getCrossSchemaResolvers = (
           })
         },
       },
-      ...quoteDetailsTable("CompleteQuote"),
-      ...quoteDisplayName("CompleteQuote")
+      ...quoteDetailsTable(getQuoteDetailsFragment("CompleteQuote")),
+      ...quoteDisplayName(getQuoteDetailsFragment("CompleteQuote"))
```
sampettersson, line 409:
```suggestion
      ...quoteDisplayName("CompleteQuote")
```

src/resolvers/cross-schema/index.ts
```diff
@@ -533,3 +533,102 @@ export const getCrossSchemaResolvers = (
...
+//     }
+//     ... on DanishTravelAgreement {
+//         street
+//         zipCode
+//         coInsured
+//     }
+//   }
+// }`
+
+function getQuoteDetailsFragment(quoteType:String): String {
```
sampettersson, line 580:
```suggestion
const getQuoteDetailsFragment = (quoteType: String) => {
```

sampettersson (changes requested):

src/resolvers/cross-schema/index.ts
```diff
@@ -324,6 +324,172 @@ export const getCrossSchemaResolvers = (
           })
         },
       },
```
hugokallstrom, line 326: Do we want these fragment definitions in `quoteDetailsTable` maybe? As `const`, `enum` or something else?

src/resolvers/cross-schema/quoteDetailsTable.ts
```diff
@@ -41,6 +40,10 @@ export const crossSchemaExtensions = `
     extend type BundledQuote {
         detailsTable(locale: Locale!): Table!
     }
+    
+    extend type Contract {
+      detailsTable(locale: Locale!): Table!
```
sampettersson, line 45: should we maybe have a `currentAgreementDetailsTable` and a `upcomingAgreementDetailsTable`? where current looks at current agreement and upcoming looks at upcomingAgreement, then its also easy to diff client-side to only show the changes.

hugokallstrom: Ok so rename `Contract` extension to `currentAgreementDetailsTable`, and `CompleteQuote` + `BundledQuote` to `upcomingAgreementDetailsTable`. Should we have those two as separate `crossSchemaExtensions` and re-use the resolve function?

sampettersson: @hugokallstrom no, not really, quotes are fine!

On contracts we have both the `Contract.currentAgreement` that we use on the contract subview to show the details + this new `ContractStatus.upcomingAgreement` that is used after the user has accepted the new quotes right?

hugokallstrom: Ah yes thats right, I'll add that!

sampettersson (approved):

src/resolvers/cross-schema/detailstable/tableRenderer.ts
```diff
@@ -0,0 +1,69 @@
+import { TableRow } from './table'
+
+export class TableRenderer {
+
+    textKeyMap: any
+    private labelMap: Record<string, string>
+    private valueRenderer: Record<string, (value: any) => string>
+
+    constructor(locale: String) {
+        this.textKeyMap = require(`${__dirname}/../../../translations/${locale}.json`)
```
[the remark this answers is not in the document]
hugokallstrom: Yes, it seems to work :)

PR #1167, opened 2021-04-30
MX-71: Enable a way to configure whether upstream schema stitching should happen or not during development.

bystam:
### What
- Introduces a new environment variable called `GRAPHQL_SCHEMA_INTROSPECTION_MODE` that defines how the upstream remote schema stitching should work.
- One of the modes, called `fault-tolerant`, enables fetching the schemas you have access to, but not failing if any of them are inaccessible
- Cross-schema extensions are resolved dynamically based on which upstreams were successful

[MX-71]: https://hedvig.atlassian.net/browse/MX-71

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -3,7 +3,7 @@ const PORT = Number(process.env.PORT) || 4000
 const UPSTREAM_MODE = process.env.UPSTREAM_MODE || 'remote'
 const LOCAL_MEMBERID_OVERRIDE: string | undefined = process.env.LOCAL_MEMBERID_OVERRIDE
 const GRAPHQL_SCHEMA_INTROSPECTION_MODE = process.env.NODE_ENV === 'development'
-  ? process.env.GRAPHQL_SCHEMA_INTROSPECTION_MODE || 'full'
+  ? process.env.GRAPHQL_SCHEMA_INTROSPECTION_MODE || 'fault-tolerant'
   : 'full'
 
 const PLAYGROUND_ENABLED =
--- a/src/config.ts
+++ b/src/config.ts
@@ -2,7 +2,10 @@ const BASE_URL = process.env.BASE_URL || 'https://gateway.test.hedvig.com'
 const PORT = Number(process.env.PORT) || 4000
 const UPSTREAM_MODE = process.env.UPSTREAM_MODE || 'remote'
 const LOCAL_MEMBERID_OVERRIDE: string | undefined = process.env.LOCAL_MEMBERID_OVERRIDE
-const GRAPHQL_SCHEMA_INTROSPECTION_MODE = process.env.GRAPHQL_SCHEMA_INTROSPECTION_MODE || 'full'
+const GRAPHQL_SCHEMA_INTROSPECTION_MODE = process.env.NODE_ENV === 'development'
+  ? process.env.GRAPHQL_SCHEMA_INTROSPECTION_MODE || 'full'
+  : 'full'
+
 const PLAYGROUND_ENABLED =
   (process.env.PLAYGROUND_ENABLED &&
     process.env.PLAYGROUND_ENABLED === 'true') ||
--- a/src/config.ts
+++ b/src/config.ts
@@ -2,6 +2,7 @@ const BASE_URL = process.env.BASE_URL || 'https://gateway.test.hedvig.com'
 const PORT = Number(process.env.PORT) || 4000
 const UPSTREAM_MODE = process.env.UPSTREAM_MODE || 'remote'
 const LOCAL_MEMBERID_OVERRIDE: string | undefined = process.env.LOCAL_MEMBERID_OVERRIDE
+const GRAPHQL_SCHEMA_INTROSPECTION_MODE = process.env.GRAPHQL_SCHEMA_INTROSPECTION_MODE || 'full'
 const PLAYGROUND_ENABLED =
   (process.env.PLAYGROUND_ENABLED &&
     process.env.PLAYGROUND_ENABLED === 'true') ||
@@ -48,5 +49,6 @@ export {
   SENTRY_ENV,
   GOOGLE_CLOUD,
   ANGEL_URL,
-  AVAILABLE_LOCALES
+  AVAILABLE_LOCALES,
+  GRAPHQL_SCHEMA_INTROSPECTION_MODE,
 }
--- a/src/schema.ts
+++ b/src/schema.ts
@@ -49,46 +49,6 @@ const redis = config.REDIS_CLUSTER_MODE
   : new Redis({ host: config.REDIS_HOSTNAME, port: config.REDIS_PORT })
 
 const makeSchema = async () => {
-  logger.info('Initializing schema')
-  const translationsLink = createHttpLink({
-    uri: 'https://api-eu-central-1.graphcms.com/v2/cjmawd9hw036a01cuzmjhplka/master',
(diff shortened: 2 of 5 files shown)


src/schema.ts
```diff
@@ -118,81 +118,120 @@ const makeSchema = async () => {
...
-  let apiGatewaySchema: GraphQLSchema | undefined
-  logger.info('Introspecting API Gateway schema')
-  apiGatewaySchema = makeRemoteExecutableSchema({
-    schema: await introspectSchema(apiGatewayLink),
-    link: apiGatewayLink,
-  })
-  logger.info('API Gateway schema introspected')
+  const introspectRemoteSchema = (
+    name: string,
+    source: { url?: string, authorized?: boolean, link?: ApolloLink }
```
bystam, line 123: Not 100% happy with the `source` input here, but I couldn't find a way to make it concise and simple in any other non-overengineered fashion.

bystam: Most services used a link that was based upon `authorizationContextLink`, but some needed the `optionalAuthorizationContextLink`. `lookup-service` also seemed to actually have a websocket-based link (is that something we still use?) and therefore I needed the custom `link?: AppoloLink` one

src/schema.ts
```diff
@@ -204,85 +243,19 @@ const makeSchema = async () => {
...
-      ...(resolvers as IResolvers<any, Context>),
-    },
-  })
-  logger.info('AppContentService Introspected')
+  const extensions: string | undefined = config.GPAPHQL_SCHEMA_INTROSPECTION_MODE !== "none"
+    ? crossSchemaExtensions
+    : undefined
+  const extensionResolvers: IResolvers | undefined = config.GPAPHQL_SCHEMA_INTROSPECTION_MODE !== "none"
+    ? getCrossSchemaResolvers(transformedTranslationSchema, contentServiceSchema!)
+    : undefined
```
bystam, line 258: Not 100% sure this is the right solution. It's also not obvious how one would go about handling these when running `mode: fault-tolerant`, as we don't know whether schemas were successfully stitched or not.

src/schema.ts
```diff
@@ -118,81 +118,120 @@ const makeSchema = async () => {
...
-  )
-  let paymentServiceSchema: GraphQLSchema | undefined
-  logger.info('Introspecting PaymentServiceSchema')
-  paymentServiceSchema = makeRemoteExecutableSchema({
-    schema: await introspectSchema(paymentServiceLink),
-    link: paymentServiceLink,
-  })
-  logger.info('PaymentServiceSchema Introspected')
+    switch (config.GPAPHQL_SCHEMA_INTROSPECTION_MODE) {
+      case "none": return undefined
```
hedvigoscar, line 144: Should we forbid these modes when running in production/staging? I think they are super-good for development, but may cause issues when running the services in production.

src/schema.ts
```diff
@@ -118,81 +118,120 @@ const makeSchema = async () => {
...
   })
-  logger.info('ProductPricingServiceSchema Introspected')
-
-  const accountServiceLink = authorizationContextLink.concat(
-    createHttpLink({
-      uri: process.env.ACCOUNT_SERVICE_GRAPHQL_ENDPOINT,
-      fetch: fetch as any,
-      credentials: 'include',
-    }),
+  const apiGatewaySchema = await introspectRemoteSchema(
```
hedvigoscar, line 165: I think we can probably await all of these in parallel (i.e. `await Promise.all[...]`) - that would be a cool improvement to add!

hedvigoscar (changes requested):
Seeing `giraffe` being shown some proper love again brings tears to my eyes. Bravo!

src/config.ts
```diff
@@ -2,6 +2,7 @@ const BASE_URL = process.env.BASE_URL || 'https://gateway.test.hedvig.com'
 const PORT = Number(process.env.PORT) || 4000
 const UPSTREAM_MODE = process.env.UPSTREAM_MODE || 'remote'
 const LOCAL_MEMBERID_OVERRIDE: string | undefined = process.env.LOCAL_MEMBERID_OVERRIDE
+const GPAPHQL_SCHEMA_INTROSPECTION_MODE = process.env.GPAPHQL_SCHEMA_INTROSPECTION_MODE || 'full'
```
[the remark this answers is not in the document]
bystam: Oh, yes! I forgot about that.

src/schema.ts
bystam: What's our current appraoch for detecting if we run in production?

src/schema.ts
bystam: That's a very good point. The only risk I see is that destructuring them might cause accidental mixups when there are so many entries in the array, and when they are order sensitive. Especially since for instance `contentServiceSchema` is referenced for more than just the `mergeSchemas` call.

src/schema.ts
hedvigoscar: `NODE_ENV` is the usual manner to check, I think 🙂

src/config.ts
```diff
@@ -48,5 +49,6 @@ export {
   SENTRY_ENV,
   GOOGLE_CLOUD,
   ANGEL_URL,
-  AVAILABLE_LOCALES
+  AVAILABLE_LOCALES,
+  GPAPHQL_SCHEMA_INTROSPECTION_MODE,
```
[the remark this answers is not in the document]
bystam: I literally had to stare at this for a full minute to see the spelling error.

src/schema.ts
hedvigoscar: Yes, I agree, that might cause issues. Maybe we can introspect the 'important ones' separately and name them, and let the others even be anonymous in an array? We can spread them into the final schema array later on.

Something like this:

```ts
const introspectedSchemas = await Promise.all([
	introspectRemoteSchema(),
	introspectRemoteSchema(),
	introspectRemoteSchema(),
	introspectRemoteSchema(),
])

// ...

mergeSchemas({
  schemas: [
    namedSchema1,
    namedSchema2,
    ...introspectedSchemas,
  ]
})
```

bystam:
I am contemplating removing the `fault-tolerant` mode altogether. It looks fine on paper, but won't work in practice due to the nature of our cross-schema extension declarations. There is no way in which parts of the introspections can fail reliably, and the cross schema extensions then also work.

Making that somehow work reliably means doing a lot more engineering in that department, probably somehow grouping extensions on which combinations of upstream sources they depend on, and dynamically resolving whether they should apply or not based on whether their dependencies failed to be introspected or not.

hedvigoscar:
I think an approach where we split up the cross-schema extensions per their dependencies is not entirely unthinkable, in particular as it would enable the 'best possible' developer experience.

Something like where fault-tolerant loads cross-schema extensions selectively if all of its dependencies are satisifed, sort of. 🙂

bystam:
> I think an approach where we split up the cross-schema extensions per their dependencies is not entirely unthinkable, in particular as it would enable the 'best possible' developer experience.
> 
> Something like where fault-tolerant loads cross-schema extensions selectively if all of its dependencies are satisifed, sort of. 🙂

Would you have time to help me pull that off? I think that might require two pairs of eyes to go through the `cross-schema` files.

hedvigoscar:
Let's see if we can make it happen this afternoon!

src/schema.ts
hedvigoscar: I really don't know about the lookup service-link, I was never involved with that service 😧

PR #1173, opened 2021-05-03
Implement Contract Bundles in Giraffe, including a move from product-pricing to Giraffe for the contract query

bystam:
### What
- Re-implement the Contracts API inside Giraffe instead of inside product-pricing
- Also implement `ContractBundle` as a type that groups contracts in bundles
- Re-imagine the `SelfChangeEligibility` concept to be local to one of these bundle types, as its `angelStories` property
- Includes a solution to [MX-117] as well

### Why
- Basically me putting my money where my mouth is, with relation to this: https://www.notion.so/hedviginsurance/A-case-for-exposing-APIs-to-Giraffe-6ac290ac76934ab588e87da45870d104
- This will enable us to make `product-pricing` have a "low level" image of what contracts are, and let `giraffe` provide the high level one, which is more in line with what is expected in the user experience.
- This also will help solving a case in the Avy partnership, where we need to "fake" the existance of a contract based on non-contract related data. That can easily be implemented in this resolver.

#### Sweden
image
#### Norway
image

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [x] Codescouted

[MX-94]: https://hedvig.atlassian.net/browse/MX-94

[MX-117]: https://hedvig.atlassian.net/browse/MX-117


src/resolvers/contracts.test.ts
```diff
@@ -0,0 +1,706 @@
...
+  ...baseOutput,
+  typeOfContract: TypeOfContract.DK_ACCIDENT,
+  displayName: 'CONTRACT_DISPLAY_NAME_DK_ACCIDENT',
+  currentAgreement: {
+    ...baseOutputAgreement,
+    address: address,
+    numberCoInsured: 2,
+    type: DanishAccidentLineOfBusiness.REGULAR,
+  },
+}
```
bystam, line 706: I would've loved to rewrite this file to actually start an apolloserver with mocked upstreams and then just fire real GraphQL towards it. Maybe even apply snapshot testing instead of comparing to explicit objects. But I will consider that a part 2 to this.

src/schema.graphqls
```diff
@@ -783,3 +686,402 @@ type File {
...
+  area: Int!
+  displayName: String!
+  hasWaterConnected: Boolean!
+}
+
+type ExtraBuildingOther implements ExtraBuildingCore {
+  area: Int!
+  displayName: String!
+  hasWaterConnected: Boolean!
+}
```
bystam, line 1069: Are all these types explicitly referenced by the client? It would be nice to maybe just collapse them to a simple type - it feels like this might be type algebra drawn to a real extreme.

src/schema.graphqls
```diff
@@ -783,3 +686,402 @@ type File {
...
+
+type InsuranceTerm {
+    displayName: String!
+    url: String!
+    type: InsuranceTermType!
+}
+
+enum InsuranceTermType {
+    TERMS_AND_CONDITIONS
+}
```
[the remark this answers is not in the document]
bystam: It works when I run it locally at least.

src/resolvers/contracts.test.ts
bystam: I should also say that I modeled these tests after what the GraphQL type mapper looked like in the backend. So it's not just made up :)

src/utils/types.ts
```diff
@@ -0,0 +1,38 @@
...
+ * }
+ * 
+ * // Solution
+ * const temp: Typenamed<Tempterature, PossibleTemperatureTypeNames> = {
+ *   __typename: 'Centrigrades', // Works!
+ *   centrigrates: 25
+ * }
+ * ``` 
+ */
+export type Typenamed<T, N> = T & { __typename: N }
```
bystam, line 38: Had our codegen instead declared the types like this:
```typescript
interface Centigrades {
  __typename?: 'Centigrades',
  centigrades: number
}

interface Farenheit {
  __typename?: 'Farenheit',
  farenheit: number
}
```

this wouldn't be needed at all. But it doesn't support it. There are alternatives that do, which we should look into using. But for now I just wanted to solve this in a well-documented manner.

src/resolvers/contracts.ts
```diff
@@ -0,0 +1,300 @@
...
+        termination: contract.terminationDate,
+      }
+  }
+}
+
+const transformAgreement = (agreement: AgreementDto): Agreement => {
+  const statusMap = {
+    [AgreementStatusDto.ACTIVE]: AgreementStatus.ACTIVE,
+    [AgreementStatusDto.ACTIVE_IN_FUTURE]: AgreementStatus.ACTIVE_IN_FUTURE,
+    [AgreementStatusDto.ACTIVE_IN_PAST]: AgreementStatus.ACTIVE,
```
bystam, line 205: Don't know why these aren't 1-to-1, but this is what it looked like in the backend:

image

src/resolvers/contracts.ts
```diff
@@ -0,0 +1,300 @@
...
+    [AgreementStatusDto.PENDING]: AgreementStatus.PENDING,
+    [AgreementStatusDto.TERMINATED]: AgreementStatus.TERMINATED,
+  }
```
bystam, line 208: Don't know why these aren't 1-to-1, but this is what it looked like in the backend:

image

src/schema.graphqls
hugokallstrom: Yeah on Android we are just querying ExtraBuildingCore, not sure why we need all these types.

src/resolvers/contracts.ts
```diff
@@ -0,0 +1,300 @@
...
+    case ContractStatusDto.TERMINATED_IN_FUTURE:
+      return {
+        __typename: 'TerminatedInFutureStatus',
+        futureTermination: contract.terminationDate,
+      }
+    case ContractStatusDto.TERMINATED:
+      return {
+        __typename: 'TerminatedStatus',
+        termination: contract.terminationDate,
+      }
```
bystam, line 197: `TerminatedTodayStatus` and `PendingStatus` are both arguably incorrectly implemented, but that's true for product-pricing as well. The concept of "now" as a date value does not exist without knowledge of the timezone, which we don't get through the headers. And `pendingSince` is computed through the `createdAt` value of the contract, which is an Instant, not a date. So that also is an inaccurate representation without a timezone.

Buuuut - my guess is these values aren't actually business critical :)

src/schema.graphqls
sampettersson: Same! Lets just 🗑️ the unnecessary types

hedvigoscar: nuke it ☢️

hedvigoscar (approved):
A bit hard to digest all the code given the magnitude of the PR, but from what I can tell: LGTM!

src/schema.graphqls
bystam: Please double check what I've done ^^ https://github.com/HedvigInsurance/giraffe/pull/1173/commits/333c3992386f14ab71cb668cda3326aef4912b67

@hugokallstrom @sampettersson maybe?

PR #1175, opened 2021-05-03
Bundle inception

sampettersson:
This creates the notion of a bundle telling the clients how it's able to control startDate and currentInsurer.

- `ConcurrentInception` - Requires all bundled quotes to have the same startDate & currentInsurer, clients should when updating startDate & currentInsurer call `editQuote` with the corresponding updates on all `correspondingQuotes.id` - This is the behaviour in DK (DK doesn't today have any defined switching process, so currentInsurer will for now always be null)
- `IndependentInceptions` - All quotes are able to fully control their startDate & currentInsurer on their own basis - This is the behaviour in SE and NO

Not 100% sure of this solution, but its for sure better than spreading this logic into 3 clients, any other solution suggestions are welcomed 🙌


src/resolvers/cross-schema/index.ts
```diff
@@ -375,7 +379,8 @@ const quotesExtension: CrossSchemaExtension = {
       },
       ...quoteDetailsTable("BundledQuote"),
       ...quoteDisplayName("BundledQuote")
-    }
+    },
+    ...quoteBundleInception(schemas)
```
bystam, line 383: Instead of attaching them here, why not make `quoteBundleInception.ts` actually export a single top level `CrossSchemaExtension` object? Feels that would make that file a bit more cohesive, since it would then also have its `dependencies` spelled out in the same file.

src/resolvers/cross-schema/quoteBundleInception.ts
```diff
@@ -0,0 +1,199 @@
...
+  correspondingQuote: {
+    __typename: string
+    id: string
+  }
+}
+
+interface IndependentInceptions {
+  __typename: string
+  inceptions: IndependentInception[]
+}
```
bystam, line 60: I really am no expert on typescript, but shouldn't/couldn't you explicitly spell out `__typename: "IndependentInceptions"` here to get typescript to help you out when constructing them?

src/resolvers/cross-schema/index.ts
```diff
@@ -16,10 +17,12 @@ export enum SchemaIdentifier {
   LOOKUP_SERVICE = "lookup-service",
 }
 
+export type Schemas = (identifier: SchemaIdentifier) => GraphQLSchema
```
bystam, line 20: 🚀

src/resolvers/cross-schema/index.ts
sampettersson: @bystam ahh, that makes more sense, updated the PR now!

src/resolvers/cross-schema/quoteBundleInceptionExtension.ts
```diff
@@ -0,0 +1,201 @@
...
+  startDate: string
+  currentInsurer: CurrentInsurer
+}
+
+interface QuoteBundle {
+  quotes: Quote[]
+}
+
+function isConcurrentInception(object: any): object is ConcurrentInception {
+  return object?.__typename == "ConcurrentInception"
```
styrelius, line 47: Very minor thing, especially given that I have a hard time understanding what's going on here more generally due to being new to all of this 😅, but just wondering if there's a reason for using `==` and not strict equality here?

hedvigoscar: If I had to guess, it's probably out of habit from Swift 😉

styrelius: Ah, I see! 😄

sampettersson: @hedvigoscar hahah very accurate observation

@styrelius fixed! 😄

hedvigoscar: I know because I've been there 😞

hugokallstrom (approved):
LGTM

PR #1187, opened 2021-05-10
Set up GraphQL based testing, where we can run real queries towards Apollo

bystam:
# Jira Issue: []

## What?
- Set up so that our Jest tests can run towards an Apollo server with the local schema
- Enables us to write tests by writing actual GraphQL queries

## Why?
- This makes us able to write tests in the same manner that the downstream clients consume them
- This is a super powerful way to prevent regressions. We can even pull queries run from the clients and write tests with those exact queries, making sure that we don't break them when making changes.

## Future work
Rewrite some tests to use [Snapshot testing](https://jestjs.io/docs/snapshot-testing).

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


package.json
```diff
@@ -56,7 +56,9 @@
     "@types/ramda": "^0.26.18",
     "@types/uuid": "^3.4.5",
     "@types/ws": "^6.0.1",
-    "apollo": "^2.16.3",
+    "apollo": "^2.21.0",
+    "apollo-server": "^2.21.0",
```
bystam, line 60: Making `apollo-server-koa` compatible with `apollo-server-testing` was a bit of a weird mess and required me to hard-lock some of them to specific versions. Therefore I decided to do what they do in their guides and simply use this one (not sure what this one actually is?)

src/schema.ts
```diff
@@ -138,25 +135,32 @@ const makeSchema = async () => {
...
+      identifier: result.identifier,
+      schema: result.schema && transformSchema(
+        result.schema,
+        [
+          new FilterRootFields(
+            (_, name) => !!allowedRootFields.find((allowedName) => name === allowedName)
+          ),
+        ]
+      )
+    })),
```
bystam, line 163: This change makes the `GRAPH_CMS` schema optional upon launch as well, when developing. Previously all of them were except for this one, because it was returned and then referenced in `member.ts`. I simply made that optional there as well for now.

In the future it would be nice to remove it from `member.ts` entirely.

src/test-utils/test-server.ts
```diff
@@ -0,0 +1,74 @@
...
+    context: context
+  })
+  const { query, mutate } = createTestClient(server)
+  return {
+    query,
+    mutate,
+    upstream,
+    stop: server.stop
+  }
+}
```
bystam, line 74: Not sure how one normally models "testing utils", but I did it like this now 🤷

package.json
```diff
@@ -56,7 +56,8 @@
     "@types/ramda": "^0.26.18",
     "@types/uuid": "^3.4.5",
     "@types/ws": "^6.0.1",
-    "apollo": "^2.16.3",
+    "apollo": "2.16.3",
```
bystam, line 59: `^2.16.3` started resolving towards a much later version (`2.33.1`) which required bumping our node version and stuff, so I decided to lock it down.

PR #1197, opened 2021-05-17
load localized strings from the files correctly

bystam:
# Jira Issue: []

## What?
- Read the correct files in the localized strings
- Add tests to verify it

## Why?
- I failed to do it last time, and there were quite a few moving pieces so I missed trying it out

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [ ] Codescouted


hedvigoscar (approved):
👏

PR #1198, opened 2021-05-17
Apply the gql tag to schema extensions, getting early schema compilation and syntax highlighting

bystam:
# Jira Issue: []

## What?
- Compile GraphQL schema extensions using `gql` to verify their syntax and get highlighting

## Why?
- Types are sometimes better than strings? 🤷 

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [x] Codescouted


hedvigoscar (approved):
🚀

PR #1203, opened 2021-05-19
MX-137: Refactor the self-change tests

bystam:
# Jira Issue: [MX-137]

## What?
- Rewrite tests for self change to instead run on a GraphQL level
- Update some testing related dependencies to make tests work better and not print any warnings

## Why?
- This PR had a lot more functionality, but that was moved to underwriter and now it's just the testing left

## Optional checklist
- [x] Unit tests written
- [ ] Tested locally
- [ ] Codescouted

[MX-137]: https://hedvig.atlassian.net/browse/MX-137


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "giraffe",
   "main": "dist/index.js",
   "scripts": {
-    "test": "jest --silent",
+    "test": "jest",
```
bystam, line 5: `--silent` is bad - we want warnings and print statements to show up. I fixed the warnings intead

package.json
```diff
@@ -60,11 +60,11 @@
     "apollo-server-testing": "2.21.0",
     "concurrently": "^4.1.1",
     "graphql-schema-typescript": "^1.5.2",
-    "jest": "^24.9.0",
+    "jest": "^26.6.3",
     "jest-expo": "^41.0.0",
     "nodemon": "^1.19.1",
     "prettier": "^1.18.2",
-    "ts-jest": "^24.2.0",
+    "ts-jest": "^26.5.6",
```
bystam, line 67: These updates enables silencing of a bunch of warnings

src/api/upstreams/underwriter.ts
```diff
@@ -69,34 +71,51 @@ export interface CreateQuoteDto {
...
       }
-      return { status: 'success', ...result}
-    }
+      return { status: 'success', ...result }
+    },
+    changeToQuotes: async (contractIds, quoteIds) => {
+      const body = { contractIds, quoteIds }
+      const result = await client.post('/quotes/selfChange', body)
+      return await result.json()
+    },
```
bystam, line 119: This file is a bit hypothetical so far, as it is not yet finalized in underwriter. This will be based on

https://github.com/HedvigInsurance/underwriter/pull/209/files

src/schema.graphqls
```diff
@@ -62,6 +62,8 @@ type Mutation {
   updatePickedLocale(pickedLocale: Locale!): Member!
   """Create all the quotes needed in relation to a change of address, based on the current state of the member's insurance."""
   createAddressChangeQuotes(input: AddressChangeInput!): AddressChangeQuoteResult!
+  """Commit the address change quotes, installing them as the ."""
```
hedvigoscar, line 65: As the what? 😧

src/schema.graphqls
```diff
@@ -62,6 +62,8 @@ type Mutation {
...
+  commitAddressChangeQuotes(contractBundleId: ID!, quoteIds: [ID!]!): AddressChangeCommitResult!
```
hedvigoscar, line 66: Do these parameters represent:

`contractBundleId`: The contract bundle to change
`quoteIds`: The new quotes to be applied to the contract bundle

?

Regardless of how correct my interpretation is, we should maybe add that as documentation on the parameter-level, just to avoid any possible confusion 🙂

src/schema.graphqls
```diff
@@ -431,6 +431,10 @@ type AddressChangeQuoteFailure {
   breachedUnderwritingGuidelines: [String!]!
 }
 
+type AddressChangeCommitResult {
+  newContractBundle: ContractBundle!
```
hedvigoscar, line 435: Are there no known failure-cases for performing this commit?

hedvigoscar (approved):
The schema looks good. I think!

src/schema.graphqls
bystam: This looks a lot like me not being used to VSCode not auto-saving... :P

hedvigoscar: I may have been burned by that as well before 😬 

Come to think of it, I should probably see if there's a setting so that I can get the same behaviour as I have in Android Studio 🤔

src/schema.graphqls
hedvigoscar: Fair enough. Our designs have a UW-limits failure case for this, but if it's not possible to reach that case from this API-call, we can probably remove it from the designs, and skip implementing it.

PR #1246, opened 2021-06-18
MX-162: wire the Trial model into 'fake contracts', to make them compatible with existing apps

bystam:
# Jira Issue: [MX-162]

## What?
- Wire the new `Trial` model onto the `Contract` queries inside Giraffe, turning them into a fake "Swedish apartment contract"
- Draft because the details are definitely not done here

## Why?
- These trials match something Avy members have been given as a "one month free" concept
- To avoid squeezing all of this into C&A, we built a new decoupled concept called 'trials'
- In order for it to be backwards compatible with the apps, we had to retrofit it onto the contracts model at some point

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [x] Codescouted

## Screenshots
image

[MX-162]: https://hedvig.atlassian.net/browse/MX-162

### DIFF
--- a/src/resolvers/contracts.ts
+++ b/src/resolvers/contracts.ts
@@ -332,16 +332,47 @@ const transformTrialToFakeContract = (trial: TrialDto, strings: LocalizedStrings
     'SE_APARTMENT_BRF': SwedishApartmentLineOfBusiness.BRF,
     'SE_APARTMENT_RENT': SwedishApartmentLineOfBusiness.RENT,
   }
-  const typeOfContract = typeTransformation[trial.type]
+  const agreementStatusTransformation: Record<string, AgreementStatus> = {
+    'ACTIVE_IN_FUTURE_AND_TERMINATED_IN_FUTURE': AgreementStatus.ACTIVE_IN_FUTURE,
+    'TERMINATED_IN_FUTURE': AgreementStatus.ACTIVE,
+    'TERMINATED_TODAY': AgreementStatus.ACTIVE,
+    'TERMINATED': AgreementStatus.TERMINATED,
+  }
+  let status: Typenamed<ContractStatus, PossibleContractStatusTypeNames>
+  if (trial.status === 'ACTIVE_IN_FUTURE_AND_TERMINATED_IN_FUTURE') {
+    status = {
+      __typename: 'ActiveInFutureAndTerminatedInFutureStatus',
+      futureInception: trial.fromDate,
+      futureTermination: trial.toDate
+    }
+  } else if (trial.status === 'TERMINATED_IN_FUTURE') {
+    status = {
+      __typename: 'TerminatedInFutureStatus',
+      futureTermination: trial.toDate
+    }
+  } else if (trial.status === 'TERMINATED_TODAY') {
+    status = {
+      __typename: 'TerminatedTodayStatus',
+      today: trial.toDate
+    }
+  } else if (trial.status === 'TERMINATED') {
+    status = {
+      __typename: 'TerminatedStatus',
+      termination: trial.toDate
+    }
+  } else {
+    status = {
+      __typename: 'ActiveStatus',
+      pastInception: trial.fromDate,
+    }
+  }
 
+  const typeOfContract = typeTransformation[trial.type]
   return {
     id: `fakecontract:${trial.id}`,
     holderMember: trial.memberId,
     typeOfContract,
-    status: {
-      __typename: 'ActiveStatus',
-      pastInception: trial.fromDate
-    } as ContractStatus,
+    status,
     displayName: strings(`CONTRACT_DISPLAY_NAME_${typeOfContract}`),
     createdAt: trial.createdAt,
     currentAgreement: {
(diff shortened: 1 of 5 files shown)


src/resolvers/contracts.ts
```diff
@@ -316,3 +322,45 @@ const moveHomeContentsToTop = (contracts: ContractDto[]) => {
...
+  const typeOfContract = typeTransformation[trial.type]
+
+  return {
+    id: `fakecontract:${trial.id}`,
+    holderMember: trial.memberId,
+    typeOfContract,
+    status: {
+      pastInception: trial.fromDate
+    },
+    displayName: strings(`CONTRACT_DISPLAY_NAME_${typeOfContract}`),
```
bystam, line 344: Mayber a cusdtom display name?

hedvigoscar: Perhaps! What have the designers said? 😄

hedvigoscar (approved):
lgtm

src/resolvers/contracts.ts
bystam: Hah! This hasn't been designed 😈

src/resolvers/contracts.ts
```diff
@@ -316,3 +322,78 @@ const moveHomeContentsToTop = (contracts: ContractDto[]) => {
...
+    'SE_APARTMENT_RENT': SwedishApartmentLineOfBusiness.RENT,
+  }
+  const agreementStatusTransformation: Record<string, AgreementStatus> = {
+    'ACTIVE_IN_FUTURE_AND_TERMINATED_IN_FUTURE': AgreementStatus.ACTIVE_IN_FUTURE,
+    'TERMINATED_IN_FUTURE': AgreementStatus.ACTIVE,
+    'TERMINATED_TODAY': AgreementStatus.ACTIVE,
+    'TERMINATED': AgreementStatus.TERMINATED,
+  }
+  let status: Typenamed<ContractStatus, PossibleContractStatusTypeNames>
+  if (trial.status === 'ACTIVE_IN_FUTURE_AND_TERMINATED_IN_FUTURE') {
```
vonElfvin, line 342: If guess we could use a `Record` here as well, would be some redundant computation but maybe the code will be more readable 🤷

src/resolvers/contracts.ts
```diff
@@ -316,3 +322,78 @@ const moveHomeContentsToTop = (contracts: ContractDto[]) => {
...
+    status = {
+      __typename: 'TerminatedStatus',
+      termination: trial.toDate
+    }
+  } else {
+    status = {
+      __typename: 'ActiveStatus',
+      pastInception: trial.fromDate,
+    }
+  }
```
vonElfvin, line 368: Code to think of it, why do we want to explode for `AgreementStatus` and not here? 🤔

src/resolvers/contracts.ts
```diff
@@ -316,3 +322,78 @@ const moveHomeContentsToTop = (contracts: ContractDto[]) => {
...
+      premium: {
+        amount: '0',
+        currency: 'SEK'
+      },
+      status: agreementStatusTransformation[trial.status],
+      address: {
+        street: trial.address.street,
+        postalCode: trial.address.zipCode,
+      },
+      squareMeters: trial.address.livingSpace ?? 0,
```
vonElfvin, line 393: Would make some sense if squareMeters was actually nullable (so we hide the UI elements appropriately), but for now I guess this will have to do 🤔 When we get generic DTOs this will most likely be nullable

vonElfvin: And by then we'll probably have some trial representation as well

src/resolvers/contracts.ts
```diff
@@ -316,3 +322,78 @@ const moveHomeContentsToTop = (contracts: ContractDto[]) => {
...
+  return {
+    id: `fakecontract:${trial.id}`,
+    holderMember: trial.memberId,
+    typeOfContract,
+    status,
+    displayName: strings(`CONTRACT_DISPLAY_NAME_${typeOfContract}`),
+    createdAt: trial.createdAt,
+    currentAgreement: {
+      __typename: 'SwedishApartmentAgreement',
+      id: `fakeagreement:${trial.id}`,
```
vonElfvin, line 380: Is Android using `UUID` for these `id`s?

src/resolvers/contracts.ts
```diff
@@ -316,3 +322,78 @@ const moveHomeContentsToTop = (contracts: ContractDto[]) => {
...
+        amount: '0',
+        currency: 'SEK'
+      },
+      status: agreementStatusTransformation[trial.status],
+      address: {
+        street: trial.address.street,
+        postalCode: trial.address.zipCode,
+      },
+      squareMeters: trial.address.livingSpace ?? 0,
+      numberCoInsured: 0,
```
vonElfvin, line 394: Same here really, showing `+ 0` in the app will yield confused members, but we have no other choice for now

vonElfvin (approved):
Nice!

src/resolvers/contracts.ts
bystam: Good point. That was just because I thought I'd try to make the variable `const` lol. That was inconsistent. I will change to Record

src/resolvers/contracts.ts
bystam: Oscar approved the PR, so I would guess not.

PR #1252, opened 2021-06-24
Quote bundle app configuration

sampettersson:
## What?
- Add AppConfiguration to quote bundle
- Fix bug in details table renderer

## Why?
- Moving flow needs this

### DIFF
--- a/src/resolvers/cross-schema/quoteBundleAppConfiguration.ts
+++ b/src/resolvers/cross-schema/quoteBundleAppConfiguration.ts
@@ -1,0 +1,54 @@
+import { gql } from 'apollo-server-koa';
+import { SchemaIdentifier, CrossSchemaExtension } from "./index"
+
+interface Quote {
+  initiatedFrom: String
+}
+
+interface QuoteBundle {
+  quotes: Quote[]
+}
+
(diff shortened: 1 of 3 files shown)


src/resolvers/cross-schema/quoteBundleAppConfiguration.ts
```diff
@@ -0,0 +1,54 @@
...
+        appConfiguration: {
+          fragment: `fragment QuoteBundleAppConfigurationCrossSchemaFragment on QuoteBundle {
+            quotes {
+              initiatedFrom
+            }
+          }`,
+          resolve: (
+            quoteBundle: QuoteBundle,
+          ) => {
+            const firstQuote = quoteBundle.quotes[0]
```
hugokallstrom, line 40: Is it enough to only check the first quote here? I suppose all quotes _should_ have the same `initiatedFrom` value though.

hugokallstrom (approved):
Nice!

src/resolvers/cross-schema/quoteBundleAppConfiguration.ts
sampettersson: @hugokallstrom yeah, I thought about that, but I have no idea what to do if they differ though?

PR #1262, opened 2021-07-05
Quote bundle displayName & gradientOption

sampettersson:
## What?

- Add displayName to QuoteBundle
- Add gradientOption to QuoteBundle

## Why?
- Needed in apps for Offer


src/resolvers/cross-schema/quoteBundleAppConfiguration.ts
```diff
@@ -17,9 +18,16 @@ export const createQuoteBundleAppConfigurationExtension = (): CrossSchemaExtensi
         UPDATE_SUMMARY
     }
 
+    enum QuoteBundleAppConfigurationGradientOption {
+        GRADIENT_ONE
```
hugokallstrom, line 22: Should we name these as they are named in the apps? Like "summer sky" etc.

hugokallstrom (approved):
LGTM

PR #1265, opened 2021-07-07
Add extension that signs or approves quotes

sampettersson:
# Jira Issue: [APP-557]

## What?

> Adds a mutation that approves a quote if its signMethod is "APPROVE_ONLY" else forwards to `signQuotes`

## Why?

> So that the apps don't need to know about the difference between approval and signing in the offer screen

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [ ] Codescouted

[APP-557]: https://hedvig.atlassian.net/browse/APP-557


hugokallstrom (approved):
LGTM!

PR #1267, opened 2021-07-08
MX-117: Sort contracts in natural order, not only by moving home content to top

bystam:
# Jira Issue: [MX-117]

## What?
- Sort all kinds of contracts in natural order, not only home content

## Why?
- That's what's in the ticket :D 

## Optional checklist
- [x] Unit tests written
- [ ] Tested locally
- [x] Codescouted

[MX-117]: https://hedvig.atlassian.net/browse/MX-117

### DIFF
--- a/src/resolvers/contracts.ts
+++ b/src/resolvers/contracts.ts
@@ -67,7 +67,7 @@ export const contracts: QueryToContractsResolver = async (
     upstream.productPricing.getMemberContracts(),
     upstream.productPricing.getTrials()
   ])
-  moveHomeContentsToTop(contracts)
+  sortContractsInNaturalOrder(contracts)
   const mappedContracts = contracts.map((c) => transformContract(c, strings))
   const fakeContracts = trials.map((t) => transformTrialToFakeContract(t, strings))
   return mappedContracts.concat(fakeContracts)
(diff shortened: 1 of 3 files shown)


src/resolvers/contracts.ts
```diff
@@ -315,12 +315,13 @@ const transformAgreement = (
   }
 }
 
-const moveHomeContentsToTop = (contracts: ContractDto[]) => {
-  contracts.sort((c1, c2) => {
-    if (c1.typeOfContract.includes('HOME_CONTENT')) return -1
-    if (c2.typeOfContract.includes('HOME_CONTENT')) return 1
-    return 0
-  })
+const sortContractsInNaturalOrder = (contracts: ContractDto[]) => {
```
sampettersson, line 318: Oh, I did this: https://github.com/HedvigInsurance/giraffe/blob/b1dfe0a6a69cb203f7ccaf07d3f28797d16caf8e/src/resolvers/cross-schema/quoteBundleOrderedQuotesExtension.ts#L58

Do you think its worth combining somehow?

bystam: Oh.. uh. I don't think it's worthwhile extracting a common function. One of them is for quotes and the other for contracts, so even if they share some commonalities they are techincally different.

However, I should probably update that function too.

bystam: PR Updated

sampettersson (approved):

PR #1276, opened 2021-07-12
Feature/claims status

anthansson:
# Jira Issue: [MX-182]

## What?
- Can fetch a member's claims filtered on enum of type ClaimStatus and ClaimOutcome

## Why?
- Allow members to follow the progress of their claims 

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [x] Codescouted

[MX-182]: https://hedvig.atlassian.net/browse/MX-182


src/resolvers/claims.ts
```diff
@@ -0,0 +1,46 @@
...
+  claims = claims.filter(o => (!_args.status || _args.status.includes(statusMap[o.status])) && // TODO refine?
+    (!_args.outcome || _args.outcome?.includes(outcomeMap[o.outcome as ClaimOutcomeDto])));
+
+  return claims.map((c) => transformClaim(c, strings))
+}
+
+const transformClaim = (
+  claim: ClaimDto, strings: LocalizedStrings
+): Claim => {
+  return <Claim>{
```
bystam, line 34: You shouldn't need this generic here `<Claim>`, and adding it actually "turns off" the typescript type checking, as it is considered a cast.

src/resolvers/claims.ts
```diff
@@ -0,0 +1,46 @@
...
+  _parent,
+  _args,
+  {upstream, strings},
+): Promise<Claim[]> => {
+  let claims : ClaimDto[] = await upstream.claimService.getMemberClaims()
+
+  claims = claims.filter(o => (!_args.status || _args.status.includes(statusMap[o.status])) && // TODO refine?
+    (!_args.outcome || _args.outcome?.includes(outcomeMap[o.outcome as ClaimOutcomeDto])));
+
+  return claims.map((c) => transformClaim(c, strings))
```
bystam, line 28: I think you can afford making this quite a bit more verbose to highlight the filtering logic

```suggestion
  const claims = await upstream.claimService.getMemberClaims()

  const filteredClaims = claims.filter(c => {
    if (_args.status) {
      if (!_args.status.includes(statusMap[c.status])) {
        return false
      }
    }
    if (_args.outcome) {
      if (!c.outcome || !_args.outcome.includes(outcomeMap[o.outcome])) {
        return false
      }
    }
    return true
  }

  return filteredClaims.map((c) => transformClaim(c, strings))
```

src/resolvers/claims.ts
```diff
@@ -0,0 +1,46 @@
...
+  [ClaimStatusDto.REOPENED]: ClaimStatus.Reopened,
+}
+
+const outcomeMap = {
+  [ClaimOutcomeDto.PAID]: ClaimOutcome.Paid,
+  [ClaimOutcomeDto.NOT_COMPENSATED]: ClaimOutcome.NotCompensated,
+  [ClaimOutcomeDto.NOT_COVERED]: ClaimOutcome.NotCompensated,
+}
+
+export const claims: QueryResolvers['claims'] = async (
```
bystam, line 18: Would love some snapshot tests for this resolver :) it should be quite easy to write

bystam (commented):
Looks nice! I had a couple of suggestions, one of them about adding a test for the claims resolver :)

src/resolvers/claims.ts
bystam: Tell me if you want to collaborate doing it!

src/resolvers/claims.ts
anthansson: Great catch! Got some errors as soon as I removed it

src/resolvers/claims.ts
anthansson: Will do!

src/resolvers/claims.ts
anthansson: Yeah I will let you know if I need any help :-) Was waiting a little until a decision has been made on the schema design

anthansson:
Will delete as I created a new branch

PR #1279, opened 2021-07-13
Move contracts, activeContractBundles query tests to snapshot testing instead, using real queries

bystam:
# Jira Issue: []

## What?
- Rewrite the contracts tests to use snapshot testing and real GraphQL queries

## Why?
- Using GraphQL for real means higher quality tests, and they are easier to maintain.
- I wanted to make sure there's a precedent for how we write tests in Giraffe

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [x] Codescouted

### DIFF
--- a/src/test-utils/test-server.ts
+++ b/src/test-utils/test-server.ts
@@ -1,4 +1,3 @@
-
 import { Upstream } from './../api/upstream';
 import { resolve } from 'path';
 import { readFileSync } from 'fs'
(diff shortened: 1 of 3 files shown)


src/test-utils/test-server.ts
```diff
@@ -77,8 +87,20 @@ export const startApolloTesting = (): TestingContext => {
...
+      }
+      return result.data!!
+    },
+    mutate: async (m) => {
+      const result = await mutate(m)
+      if (result.errors) {
+        throw result.errors
+      }
+      return result.data!!
+    },
```
bystam, line 103: This change made errors show up as thrown instead of part of an `errors` object of the return value. Throwing them means they show up way easier in the tests

hugokallstrom (approved):
LGTM!

PR #1281, opened 2021-07-14
Include more statuses in what is considered 'active'

bystam:
# Jira Issue: []

## What?
- Widen what we consider to be an 'active' contract

## Why?
- The `ContractStatus`es have a bit of overlap. A contract is "active" if it is active right now but terminated in the future, however then it won't have the `ACTIVE` status, but rather a `TERMINATED_IN_FUTURE` one.
- https://hedviginsurance.slack.com/archives/C01NB6GBRFS/p1626162624003900

## Optional checklist
- [x] Unit tests written
- [ ] Tested locally
- [x] Codescouted


hugokallstrom (approved):
LGTM!

PR #1284, opened 2021-07-15
APP-637: Address Change - allow no extra buildings when submitting for `SE_HOUSE`

hedvigoscar:
- Reformat file using prettier
- Replace `!!` with `!`
- Reformat file using prettier
- Allow undefined as input for `extraBuildings`

# Jira Issue: [APP-637]

## What?
Allow `extraBuildings` to not be submitted when creating address change quotes for
`SE_HOUSE`. If no `extraBuildings` are provided, default it to `[]` in the
upstream API.

## Why?
Some Embark-based clients (specifically android in this case) did not set any
value when no MultiAction-items were provided, leading to this mutation being
called with `extraBuildings: undefined` instead of `extraBuildings: []`. An
alternative implementation path could have been to set `extraBuildings: []`, but
we determined it to be sensible to change this mutation to instead work in this
manner.

## Optional checklist
- [X] Unit tests written
- [ ] Tested locally
- [X] Codescouted

[APP-637]: https://hedvig.atlassian.net/browse/APP-637

### DIFF
--- a/src/resolvers/__snapshots__/addressChange.test.ts.snap
+++ b/src/resolvers/__snapshots__/addressChange.test.ts.snap
@@ -189,6 +189,33 @@ Array [
 ]
 `;
 
+exports[`createAddressChangeQuotes - SE_HOUSE accepts undefined for \`extraBuildings\` 1`] = `
+Array [
+  Array [
+    Object {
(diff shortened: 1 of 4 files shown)


anthansson (approved):
Looks good to me! :-)

PR #1290, opened 2021-07-21
Add ability to create a claim via a GraphQL mutation

anthansson:
# Jira Issue: [MX-222]

## What?
- New mutation createClaim added. Adds ability to create a claim via a GraphQL mutation that communicates directly with claims-service. Previously when creating a claim the chain of communication was:  giraffe -> bot-service - claims service, since claims are created via the chat. Adding this mutation allows us to create a new workflow where members can create claims outside without using the chat. 

- Also realised that the audioUrl was not signed when fetching claims of a member, so added signing-functionality. 

- Updated the claims-service URLs and their return type 

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [x] Codescouted

[MX-222]: https://hedvig.atlassian.net/browse/MX-222


src/resolvers/claims.ts
```diff
@@ -29,6 +33,18 @@ export const claims: QueryResolvers['claims'] = async (
 const transformClaim = (
   claim: ClaimDto, _strings: LocalizedStrings
 ): Claim => {
+
+  let signedUrl = undefined
```
anthansson, line 37: The url previously not signed, so fixed that!

PR #1293, opened 2021-07-22
Add new query 'claims' + snapshot tests

anthansson:
# Jira Issue: [MX-182]

## What?
- New query 'claims' added that lets clients fetch all their claims. Initially had this [old branch](https://github.com/HedvigInsurance/giraffe/pull/1276) but created this fresh one as there was too many conflicts as we have swapped GraphQL gen library. 

## Why?
- Ultimately will allow members to follow the progress of their claims

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [x] Codescouted

[MX-182]: https://hedvig.atlassian.net/browse/MX-182


src/resolvers/claims.test.ts
```diff
@@ -0,0 +1,63 @@
...
+describe('Query.claims', () => {
+
+  it('works for zero claims', async () => {
+    upstream.claimService.getMemberClaims = () => Promise.resolve([])
+    const claims = await QUERIES.claims()
+    expect(claims).toMatchSnapshot()
+  })
+
+  it('works for claim with contract with agreement', async () => {
+    //context.upstream.claimService.createClaim = () => Promise.resolve({claimId: MEMBER_ID})
```
hedvigoscar, line 18:
```suggestion
```

src/schema.graphqls
```diff
@@ -1019,3 +1022,30 @@ type Address {
...
+  id: String!
+  contract: Contract
+  status: ClaimStatus!
+  outcome: ClaimOutcome
+  submittedAt: Instant!
+  closedAt: Instant
+  files: [File!]!
+  signedAudioURL: String
+  payout: MonetaryAmountV2
+  _contractId: ID
```
hedvigoscar, line 1037: This is an implementation detail of `giraffe` that we should not leak 🙂 
```suggestion
```

hedvigoscar (approved):
Looks mostly sensible to me. Check the comments, after they are addressed you can merge 🙂

src/schema.graphqls
anthansson: We put it there when you helped me with this.If I remove it I need to do some work arounds?

anthansson: removed 🦖

PR #1294, opened 2021-07-22
Feature/create claim mutation

anthansson:
# Jira Issue: [MX-222]

## What?
- New mutation createClaim added. Adds ability to create a claim via a GraphQL mutation that communicates directly with claims-service. Previously when creating a claim the chain of communication was:  giraffe -> bot-service - claims service, since claims are created via the chat. Adding this mutation allows us to create a new workflow where members can create claims outside without using the chat. 

- Extracted a function wrapping s3.upload in a promise which resulted in much nicer code anywhere s3.upload was used. 

- Tried to add snapshot test for creating claim and uploading file but didn't get it working. Added ticket to finish it later. 

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [x] Codescouted

[MX-222]: https://hedvig.atlassian.net/browse/MX-222


hedvigoscar:
Seems like `src/generated/graphql.ts` may have been deleted by mistake 😬

anthansson:
Got merge conflicts everywhere so I simply deleted all its content 🤪  Will generate a new one

src/schema.graphqls
```diff
@@ -31,6 +31,7 @@ type Query {
 
 type Mutation {
   logout: Boolean!
+  createClaim(audioFile: Upload!): String!
```
hedvigoscar, line 34: Canonical GraphQL-schema design would suggest that this mutation should have the following signature:

```suggestion
  createClaim(audioFile: Upload!): Claim!
```

hedvigoscar: Then, from the `createClaim` mutation-resolver, you can call the same code path that is used for resolving a single claim, to promote maximum code reuse 🙂

src/test-utils/test-server.ts
```diff
@@ -1,13 +1,13 @@
-import { Upstream } from './../api/upstream';
```
hedvigoscar, line 1: Is this file formatted with `prettier`? It looks like it may have been de-formatted 🤔

src/resolvers/createClaim.ts
```diff
@@ -0,0 +1,52 @@
...
+        Bucket: bucket,
+        Key: key,
+        Body: createReadStream(),
+        ContentType: mimetype,
+        Metadata: {
+          MemberId: user.memberId,
+        },
+      }
+
+      s3.upload(params, UPLOAD_OPTIONS, (err, data) => {
```
hedvigoscar, line 36: It would be neat to `promisify` this function, so that we can use appropriate `async`/`await`-syntax to call it, instead of callbacks.

It would probably look something like this:

```typescript
const s3Upload = (params, options) = new Promise((resolve, reject) => {
  s3.upload(params, options, (err, data) => {
    if (err) {
      reject(err)
      return
    }
    resolve(data)
  })
}
```

and we can then call it as such:

```typescript
const uploadedAudioURL = await s3Upload(...)
```

instead of having to create a weird chain of raw `Promise`-objects in our resolver function body 😄

hedvigoscar (changes requested):
Some things to look at wrt schema, and a few general improvements. Mostly good though!

src/api/upstreams/claimService.ts
```diff
@@ -43,7 +53,11 @@ export enum ClaimOutcomeDto {
 export const createClaimServiceClient = (client: HttpClient): ClaimServiceClient => {
   return {
     getMemberClaims: async () => {
-      const res = await client.get(`/claims/listClaims`)
+      const res = await client.get(`/claims`)
+      return res.json()
+    },
+    createClaim: async (createClaimDTO: CreateClaimDTO) => {
+      const res = await client.post(`/claims/createClaim`, createClaimDTO)
```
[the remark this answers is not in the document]
anthansson: Got another pr regarding this issue :-) will deploy that one before this branch

src/test-utils/test-server.ts
anthansson: true was KtLint!

src/schema.graphqls
anthansson: For anyone reading this. After a claim is created axon triggers a few events that will be handled asynchronously, thus a claim object cannot be returned right away. I will simply return the claimId for now but perhaps we gotta take a look at how to handle such inconsistencies. Feels like either the front end or back must go through some major changes? 

https://stackoverflow.com/questions/64045477/how-to-make-command-to-wait-until-all-events-triggered-against-it-are-completed/64117171#64117171

src/api/upstreams/productPricing.ts
```diff
@@ -1,4 +1,4 @@
-import {HttpClient} from '../httpClient'
+import { HttpClient } from '../httpClient'
```
anthansson, line 1: just some formatting on this file

src/resolvers/chatResponse.ts
```diff
@@ -46,30 +48,19 @@ export const sendChatAudioResponse: MutationResolvers['sendChatAudioResponse'] =
...
+    Bucket: AWS_CLAIMS_S3_BUCKET,
+    Key: key,
+    Body: createReadStream(),
+    ContentType: mimetype,
+    Metadata: {
+      MemberId: user.memberId,
     },
-  )
+  }
+  const uploadedFileUrl = await s3Upload(params, UPLOAD_OPTIONS)
```
anthansson, line 62: should I do any try-catch or just let giraffe throw errors? Will result in 500

src/resolvers/file.ts
```diff
@@ -17,19 +15,13 @@ export const fileInner = async (
...
   if (headObject.Metadata && headObject.Metadata.memberid === memberId) {
-    const signedUrl = s3.getSignedUrl('getObject', {
-      Bucket: AWS_S3_BUCKET,
-      Key: key,
-      Expires: THIRTY_MINUTES,
-    })
-
     return {
-      signedUrl,
+      signedUrl: await s3GetSignedUrl(AWS_S3_BUCKET, key),
```
anthansson, line 19: same questions in all uses of s3 functions. Should I try catch?

src/resolvers/uploadFile.ts
```diff
@@ -54,37 +47,23 @@ export const uploadFiles: MutationResolvers['uploadFiles'] = async (
...
+  const params = {
+    Bucket: AWS_S3_BUCKET,
+    Key: key,
+    Body: createReadStream(),
+    ContentType: mimetype,
+    Metadata: {
+      MemberId: memberId,
+    },
+  }
+  await s3Upload(params, UPLOAD_OPTIONS)
```
anthansson, line 62: same questions in all uses of s3 functions. Should I try catch?

src/resolvers/chatResponse.ts
hedvigoscar: Any thrown errors should result in something like this:

```json
{
	"data": null",
	errors: [
		{
			// ...
		}
	]
}
```

The primary concern would be if the error message leaks any information. Behaviourally I would argue that this is correct.

hedvigoscar (approved):
Approved, with the caveat about Eventual Consistency remaining 🙂

PR #1296, opened 2021-07-23
Add promise-cms to giraffe schema

joacimastrom:
# Jira Issue: []

## What?
- Add promise-cms to schema stiching

## Why?
- To allow clients to consume promise-cms data

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


src/schema.ts
```diff
@@ -204,6 +204,13 @@ const makeSchema = async (): Promise<SchemaIntrospectionResult> => {
         authorized: false
       }
     ),
+    introspectRemoteSchema(
+      SchemaIdentifier.PROMISE_CMS,
+      {
+        url: process.env.PROMISE_CMS_GRAPHQL_ENDPOINT,
+        authorized: false
```
anthansson, line 211: I have no idea about these things, but seems to be inline with the other stuff! Not sure whether I am "allowed" to approve or not 🤪

joacimastrom: Yeah I more or less copy+pasted the format for the other ones :) 

About authorized, I think it doesn't matter all too much atm since we're only going to test it in staging at first.

@bystam will there be any issues if this is released to prod, if we don't have an env variable for `PROMISE_CMS_GRAPHQL_ENDPOINT` in prod?

hedvigoscar: I'm going to assume that nothing good can come from that 😅 if you look at https://github.com/HedvigInsurance/giraffe/blob/5b1ca785a5c748adb2b340ef7d19758ae9f3f5ba/src/schema.ts#L95 you can see that we're actually casting it to be non-null later. Better have it supplied pre-deploy, probably!

joacimastrom: Hmm yeah that doesn't look very good... What do you mean supply it pre-deploy? If we just want to test this in staging for now, how would you proceed?

I'm thinking a simply null-check would suffice?

hedvigoscar: That's probably appropriate - something like

```typescript
process.env.PROMISE_CMS_GRAPHQL_ENDPOINT && introspectRemoteSchema(/* the rest of the code */),
```

is likely enough, given that `Promise.all` accepts undefined in its array

hedvigoscar (approved):

PR #1303, opened 2021-07-29
Add some changes to claims query API

anthansson:
# Jira Issue: [MX-182, MX-224]

## What?
- Remove private field from claims in schema
- Add url signing functionality to claim files
- Also look for trials when fetching the contract to include in a claim

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [x] Codescouted


src/resolvers/claims.ts
```diff
@@ -1,23 +1,34 @@
...
 
-const outcomeMap = {
-  [ClaimOutcomeDto.PAID]: ClaimOutcome.Paid,
-  [ClaimOutcomeDto.NOT_COMPENSATED]: ClaimOutcome.NotCompensated,
-  [ClaimOutcomeDto.NOT_COVERED]: ClaimOutcome.NotCompensated,
+export const getContractByClaim: ClaimResolvers['contract'] = async (parent, _args, context) => {
+
+  // try find contract
+  const parentAsPrivate = (parent as ClaimPrivate)
+  if (!parentAsPrivate._contractId) return undefined
```
hedvigoscar, line 18:
```suggestion
  if (!parentAsPrivate._contractId) {
    return undefined
  }
```

`if`-statements without curly braces are error-prone: https://embeddedgurus.com/barr-code/2014/03/apples-gotofail-ssl-security-bug-was-easily-preventable/

The following form works if we want to keep it as a one-liner, but really we should just go with whatever formatting `prettier` considers to be correct 🙂 

```suggestion
  if (!parentAsPrivate._contractId) { return undefined }
```

src/resolvers/claims.ts
```diff
@@ -1,23 +1,34 @@
...
+  const contract = await context.upstream.productPricing.getContract(parentAsPrivate._contractId)
+  if (contract) return transformContract(contract, context.strings)
+
+  // else try find trial
+  const trials = await context.upstream.productPricing.getTrials()
+  const allHeaders = (context.headers as any) as { [key: string]: string }
+  const memberId = allHeaders["Hedvig.token"]
+  if (!memberId) return undefined
```
hedvigoscar, line 26:
```suggestion
  if (!memberId) {
    return undefined
  }
```

src/resolvers/claims.ts
```diff
@@ -1,23 +1,34 @@
...
-  [ClaimOutcomeDto.PAID]: ClaimOutcome.Paid,
-  [ClaimOutcomeDto.NOT_COMPENSATED]: ClaimOutcome.NotCompensated,
-  [ClaimOutcomeDto.NOT_COVERED]: ClaimOutcome.NotCompensated,
+export const getContractByClaim: ClaimResolvers['contract'] = async (parent, _args, context) => {
+
+  // try find contract
+  const parentAsPrivate = (parent as ClaimPrivate)
+  if (!parentAsPrivate._contractId) return undefined
+  const contract = await context.upstream.productPricing.getContract(parentAsPrivate._contractId)
+  if (contract) return transformContract(contract, context.strings)
```
hedvigoscar, line 20:
```suggestion
  if (contract) {
    return transformContract(contract, context.strings)
  }
```

src/resolvers/claims.ts
```diff
@@ -1,23 +1,34 @@
...
+
+  // else try find trial
+  const trials = await context.upstream.productPricing.getTrials()
+  const allHeaders = (context.headers as any) as { [key: string]: string }
+  const memberId = allHeaders["Hedvig.token"]
+  if (!memberId) return undefined
+  const filteredTrials = trials.filter((c) => c.memberId == memberId)
+  const fakeContract = filteredTrials.pop()
+  if (fakeContract) return transformTrialToFakeContract(fakeContract, context.strings)
```
hedvigoscar, line 29:
```suggestion
  if (fakeContract) {
    return transformTrialToFakeContract(fakeContract, context.strings)
  }
```

src/resolvers/claims.ts
```diff
@@ -30,20 +41,34 @@ export const claims: QueryResolvers['claims'] = async (
...
+  Key: key,
+  Expires: THIRTY_MINUTES,
+})
+
+const signAudioUrl = (audioUrl: String): String => {
+  const split: Array<String> = audioUrl?.split("/")
+  const key = split[split.length - 1]
+  // console.log("CLAIM ID: ", claim.id)
+  // console.log("SIGN BUCKET: ", AWS_CLAIMS_S3_BUCKET)
+  // console.log("SIGN KEY:", key)
```
hedvigoscar, line 56:
```suggestion
```

src/resolvers/claims.ts
```diff
@@ -30,20 +41,34 @@ export const claims: QueryResolvers['claims'] = async (
   return claims.map((c) => transformClaim(c, strings))
 }
 
+
+const signURL = (bucket: String, key: String) => s3.getSignedUrl('getObject', {
+  Bucket: bucket,
+  Key: key,
+  Expires: THIRTY_MINUTES,
+})
```
hedvigoscar, line 49: Might be wise to use `getSignedUrlPromise` (https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#getSignedUrlPromise-property)

```suggestion
const signURL = (bucket: String, key: String) => s3.getSignedUrlPromise('getObject', {
  Bucket: bucket,
  Key: key,
  Expires: THIRTY_MINUTES,
})
```

This change would change the signature of this from `string` to `Promise<string>`, which means that callers would need to `await` the call to `signURL`.

hedvigoscar (changes requested):
Looks mostly sensible! A few issues to address, nothing crazy though 😄

src/resolvers/claims.ts
anthansson: getSignedUrlPromise does not seem exist in the version of the aws-sdk we are using. So I'll skip that for now? getSignedUrl is being used in more places as well.

hedvigoscar: I see! We can trivially create our own by wrapping it:

```typescript
const getSignedUrl = new Promise((resolve, reject) => {
  s3.getSignedUrl(
    'getObject',
	{
	  Bucket: bucket,
      Key: key,
  	  Expires: THIRTY_MINUTES,
  	},
    (err, url) => {
      if (err) {
		reject(err)
        return
	  }
      resolve(url)
    }
  )
})
```

hedvigoscar (approved):

PR #1305, opened 2021-07-30
fix issue where requests to localhost would not contain remotePathPrefix

anthansson:
Fix issue where requests to localhost would not contain remotePathPrefix. For example any requests made to claimsService has a remotePathPrefix = '/claims'. This would not be added as a prefix if running on localhost, hence the actual requests had to have different requests URLs depending on if running on localhost. 

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [x] Codescouted


src/api/upstream.ts
```diff
@@ -33,7 +33,7 @@ export const createUpstream = (
     const allHeaders = (headers as any) as { [key: string]: string }
 
     const baseUrl: string = config.UPSTREAM_MODE === "local"
-      ? `http://localhost:${localPort}`
+      ? `http://localhost:${localPort}${remotePathPrefix}`
```
anthansson, line 36: we want remotePathPrefix no matter what

src/api/upstreams/claimService.ts
```diff
@@ -43,7 +43,7 @@ export enum ClaimOutcomeDto {
 export const createClaimServiceClient = (client: HttpClient): ClaimServiceClient => {
   return {
     getMemberClaims: async () => {
-      const res = await client.get(`/claims/listClaims`)
+      const res = await client.get(``)
```
anthansson, line 46: Then we don't need prefix here now. The listClaims was removed in claimsService as a get request made to '/claims' is quite intuitive.

AlexSJ1990: won't we still need the /claims here?

AlexSJ1990 (commented):
Nice! one small question

src/api/upstreams/claimService.ts
anthansson: No that’s the point of the whole pr. The prefix is automatically added in upstream.ts above :)

anthansson: Just happens that ’/listClaims’ also was removed, but separetely in another pr that I haven’t merged yet. So wasn’t sure to leave or remove /listClaims here as well.

AlexSJ1990: ahh okay i see! awesome :)

AlexSJ1990 (approved):
✅

PR #1308, opened 2021-08-03
Hide null values from table

sampettersson:
## What?
- Hide null values from tables

## Why?
- We show "City" -> "null" in the table right now


src/resolvers/cross-schema/detailstable/tableRenderer.ts
```diff
@@ -47,7 +47,7 @@ export class TableRenderer {
     public mapDetailRows = (query: any): TableRow[] => Object.keys(query).map(key => {
         const labelMapValue = this.labelMap[key]
 
-        if (!labelMapValue) {
+        if (!labelMapValue || !query[key]) {
```
robinandeer, line 50: this will be true if `query[key]` is `0` as well - will that ever happen?

sampettersson: Hmm, idk, but I guess its best to do an explicit check, updated!

robinandeer (approved):
Looks good!

PR #1314, opened 2021-08-11
Feature/promise cms integration

joacimastrom:
# Jira Issue: [IPC-177], [IPC-178], [IPC-179]

## What?
- Initial integration between giraffe and promise-cms to serve perilCollections to clients
- Query for perils from promise-cms is `PerilsV2` until tested thourougly. Migration from current Perils will come at a later stage

## Why?
- Perils are today hard-coded files in content-service, meaning any additions or edits to existing perils have to go through developers and a deployment process
- Icons for perils are also stored as files in content-service. The promise-cms will instead store them in an S3-bucket and allow editors to switch them out in the administration panel

## Optional checklist
- [ ] Unit tests written
- [X] Tested locally
- [X] Codescouted

[IPC-177]: https://hedvig.atlassian.net/browse/IPC-177
[IPC-178]: https://hedvig.atlassian.net/browse/IPC-178
[IPC-179]: https://hedvig.atlassian.net/browse/IPC-179


src/schema.graphqls
```diff
@@ -579,6 +581,30 @@ type Member {
...
+
+type PerilV3 {
+  title: String
+  description: String
+  info: String
+  shortDescription: String
+  covered: [Coverage]
+  exceptions: [Coverage]
+  iconSvg: Media
+  iconPdf: Media
```
sampettersson, line 597: This needs to be of type `Icon` to be compatible with our iOS renderer, all icons need to be supplied in two variants, dark and light.

src/schema.graphqls
```diff
@@ -579,6 +581,30 @@ type Member {
   features: [Feature!]!
 }
 
+type PerilCollection {
+  perils: [PerilV3!]!
+  typeOfContract: String!
+}
+
+type PerilV3 {
```
sampettersson, line 589: Why do we need a third version? Can't we be API compatible with V2, or what's new?

joacimastrom: The datamodel created with Strapi differs a little bit but not a lot. Most notably the icons, but also the `covered/exceptions` will be an array objects as `{ description: "....." }` instead of just an array of strings

I can of course format this in giraffe to make it fit the PerilV2 type, but I'm not sure which will cause more work down the line. Since PerilV2 is defined in content-service, which we are aiming to move away from, we will have to move the type definition into giraffe

src/schema.graphqls
joacimastrom: Great, I'll adjust the format to fit the Icon type

src/schema.graphqls
robinandeer: How about `BundledQuote.perils`, `Contract.perils`, `CompleteQuote.perils`?

Those are still defined as `PerilV2`, no? I guess that peril information will need to be fetched from the CMS as well. If that API changes we will have to do more work on our end. But if Giraffe can handle the transformation, it would be nice.

joacimastrom: Yeah good point Robin. I've looked into those as well and will switch their resolvers over to this new one once we've been able to test it end to end, and I'll migrate the PerilV2-type to giraffe and format the data so clients receive the perils unchanged :)

src/schema.graphqls
```diff
@@ -27,6 +27,9 @@ type Query {
 
   """ Returns all claims the member has """
   claims: [Claim!]!
+
+  """ Returns perils from promise-cms """
+  perilsV2(typeOfContract: TypeOfContract!, locale: Locale!): [PerilV2]
```
robinandeer, line 32: Should this not be typed like this?

```suggestion
  perilsV2(typeOfContract: TypeOfContract!, locale: Locale!): [PerilV2!]!
```

joacimastrom: Does every type of contract exist for every locale? In that case yes :)

joacimastrom: I guess we wont have `typeOfContract: SE_APARTMENT_BRF, locale: nb_NO`, but maybe we want to throw an error in that case? @robinandeer

robinandeer: Right, wouldn't those just be filtered out so we end up with an empty array? I think that would be okay to deal with on the frontend, no?

```json
{
	"perilsV2": []
}
```

I believe with this typing, a valid response could look like this:

```json
{
	"perilsV2": [{ ... }, null, null, { ... }]
}
```

...and we don't want to handle that every time hehe

joacimastrom:
@robinandeer @sampettersson @hedvigoscar Do you have any more feedback on this or can we merge it and start testing? :) It shouldn't affect any existing functionality

src/api/upstreams/promiseCms.ts
```diff
@@ -0,0 +1,111 @@
...
+    dark: PromiseCmsIconVariant
+  }
+}
+
+type coverage = {
+  description: string
+}
+
+type PromiseCmsIconVariant = {
+  iconSvg: Media
```
AlexSJ1990, line 35: why not just string?

AlexSJ1990 (approved):
Looks good!

src/api/upstreams/promiseCms.ts
joacimastrom: Ah! Strapi has built in support for uploading images/videos/gifs... these are stored as with the type `Media` which includes a bunch of other attributes like `name, size, width, height...` etc

So instead of uploading an image and then adding the URL manually, we simply add the image and Strapi sorts out the rest

joacimastrom: So there is a lot more information on the `Media`-type but we only care about the `url` in this specific case

PR #1319, opened 2021-08-19
Feature/add test create claim

anthansson:
# Jira Issue: [MX-224]

Added a snapshot test for createClaim mutation 

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [x] Codescouted

[MX-224]: https://hedvig.atlassian.net/browse/MX-224


src/resolvers/__snapshots__/createClaim.test.ts.snap
```diff
@@ -0,0 +1,7 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Mutation.createClaim create claim works 1`] = `
+Object {
+  "createClaim": "c1",
+}
+`;
```
bystam, line 7: Instead of snapshotting the output (which will just be what you mocked on `upstream.claimService.createClaim`), I think we should snapshot what *input* was sent to the upstream calls, because that's where most of the actual resolver logic lies.

If you check the tests for `createAddressChangeQuotes`, you will notice that I don't test the output of the mutation, because it's rather insignificant. Instead, I mock the upstream call `createQuote` by using `jest.fn()`, and then I leverage the recorded `.calls` property for the snapshots, because that's where the interesting logic happens.

bystam (commented):
Nice!! I added some thoughts to it

PR #1323, opened 2021-08-24
Add app retirement query

tobiasbexelius:
# Jira Issue: [MX-249]

## What?
- A query to determine if the user's app needs to be updated
- If `X-Build-Version: -1` is sent, an example object is returned for testing purposes, otherwise `null`

## Why?
- So that we can ship features with breaking API changes without having the app behave badly

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [x] Codescouted

[MX-249]: https://hedvig.atlassian.net/browse/MX-249


src/resolvers/appVersionRetirement.ts
```diff
@@ -0,0 +1,22 @@
...
+) => {
+  const buildVersion = parseInt(headers["X-Build-Version"] ?? '0', 10)
+  if (buildVersion === -1) {
+    return {
+      message: "You need to update",
+      storeLink: "https://appstore.com/hedvig",
+      supportPhoneNumber: "0701234567",
+      supportEmail: "support@hedvig.com"
+    }
+    return undefined
```
rasmusguterstam, line 16: Minor, but perhaps return `null` here instead? Or is there any specific reason as to why `undefined`? The test seems to test for `null`, and I assume it works because Apollo interprets `undefined` as `null`, but I'd say it's better to be explicit.

src/resolvers/appVersionRetirement.ts
```diff
@@ -0,0 +1,22 @@
...
+  }
+}
```
rasmusguterstam, line 18:
```suggestion
const appVersionRetirement: QueryResolvers['appVersionRetirement'] = async (
  _parent,
  _args,
  {headers},
) => {
  const buildVersion = parseInt(headers["X-Build-Version"] ?? '0', 10)
  if (buildVersion === -1) {
    return {
      message: "You need to update",
      storeLink: "https://appstore.com/hedvig",
      supportPhoneNumber: "0701234567",
      supportEmail: "support@hedvig.com"
    }
  }
  
  return null
}
```

src/resolvers/appVersionRetirement.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export {
+  appVersionRetirement
+}
```
rasmusguterstam, line 22:
```javascript
export const appVersionRetirement: QueryResolvers['appVersionRetirement'] = ...
```

src/test-utils/test-server.ts
```diff
@@ -39,19 +39,19 @@ const fakeUpstream = (): Upstream => ({
...
 })
 
-const upstream = fakeUpstream()
+const defaultHeaders = () => ({
+  'User-Agent': 'Testing',
+  'X-Forwarded-For': 'Testing',
+  'X-Request-Id': 'Testing',
+  'Accept-Language': 'en',
+  'Enable-Simple-Sign': 'false',
+})
```
rasmusguterstam, line 48:
```suggestion
const defaultHeaders = ({
  'User-Agent': 'Testing',
  'X-Forwarded-For': 'Testing',
  'X-Request-Id': 'Testing',
  'Accept-Language': 'en',
  'Enable-Simple-Sign': 'false',
})
```

rasmusguterstam: And if Typescript complains, go for something like

```javascript
const defaultHeaders: Record<string, string> = {
  'User-Agent': 'Testing',
  'X-Forwarded-For': 'Testing',
  'X-Request-Id': 'Testing',
  'Accept-Language': 'en',
  'Enable-Simple-Sign': 'false',
}
```

rasmusguterstam (approved):
Nice!! Some minor suggestions, but nothing critical. Would also recommend setting up Prettier with IntelliJ and run it once (and in the future, when you do more stuff with javascript)!

src/resolvers/appVersionRetirement.ts
bystam: This is because, when we moved to the new GraphQL-to-TS generation library, the `Maybe<T>` type was defined as `= T | null`, which apparently was the standard.

However, that definition caused a LOT of errors in our existing code, as we were using `undefined` either implicitly or explicitly in lots of places. So I changed the definition of Maybe to be `T | undefined` to be able to move forward.

This also has the effect of not being able to return null for `Maybe`s. If we feel this is ugly then we could restore the original definition of Maybe and refactor the existing code.

bystam: My overridden definition exists in `codegen.yml`

src/test-utils/test-server.ts
bystam: I don't think this will work though. The reason it's a function that returns an object is because the tests mutate the object, and we want to restore it to its original state between tests. Therefore we need to copy data from an untouched version of it.

bystam: We can of course do that differently, but simply applying your suggested diff alone will probably break the code.

joacimastrom: I'd say either rename the function to ```getDefaultHeaders``` or change it as Rasmus suggested but clone it when using it instead, as ```headers: { ...defaultHeaders }```

bystam: `headers: { ...defaultHeaders }` will work so long as we don't have a deeper object structure, as it does not perform a deep copy.

We basically copied the existing case in the same file where the "mock upstream" is re-created entirely between calls, to make sure that it becomes a deep re-creation of the entire object.

But for headers it should be enough with the shallow copy, of course.

tobiasbexelius:
We went with another approach to fullfil this need :)

PR #1326, opened 2021-08-30
MX-249: a version of the 'old client' retirement feature as a Koa middleware instead of custom query

bystam:
# Jira Issue: [MX-249]

## What?
- Add Koa middleware that returns a known error format if the calling version is deprecated
- Has a "build version == -1" case for testing

## Why?
- We want the ability to tell clients that they need to update because we have made breaking changes for old APIs

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [x] Codescouted

[MX-249]: https://hedvig.atlassian.net/browse/MX-249


sampettersson (approved):
I think this solution might work best for our future use-cases.

src/middlewares/versionCheck.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * A middleware that reads the version of the calling client, and if the version headers
+ * indicate that it's a client we no longer support we throw an invalid version error.
+ */
+export const versionCheckMiddleware: Application.Middleware = async (ctx, next) => {
+    const error = versionCheck(ctx)
+    if (error) {
+        const response: ErrorResponse<InvalidVersionError> = { error }
+        ctx.response.status = 403
```
hedvigoscar, line 16: This might have unintended consequences - GraphQL-servers generally respond with status code 200, no matter if the response is successful or not. Error-handling is done by checking the `errors`-field of the response.

If 200 is not returned, there is a non-zero chance that our clients will short-circuit and never reach our application middlewares.

bystam: Good catch! Updated

hedvigoscar (approved):

PR #1327, opened 2021-08-30
Added SE Accident schemas

mightyost:
# Jira Issue: [IPC-224]

Add SE Accident agreement to PP contract schema.

Note: The bundling logic needs to be changed to support markets with more than one insurance and any types of add-ons. Right now is the contract test `'becomes two bundles for two swedish contracts'` in `contracts.test.ts` failing (and is disabled).

This PR is with a warm hand donated to App / Web team, better if you who "owns" Giraffe (and is more comfortable with typescript) implements this. [@sampettersson @gustaveen]: https://hedviginsurance.slack.com/archives/C016KUN61U3/p1630320537004700

[IPC-224]: https://hedvig.atlassian.net/browse/IPC-224


src/resolvers/contracts.ts
```diff
@@ -192,6 +194,11 @@ export const bundleContracts = (
       case 'DanishAccident':
         danishBundle.push(contract)
         break
+      case 'SwedishHouse':
+      case 'SwedishApartment':
+      case 'SwedishAccident':
+        swedishBundle.push(contract)
+        break
```
bystam, line 201: I don't think this is the correct logic here. Sweden as specifically omitted from the code because "house" and "apartment" contracts shouldn't end up in the same bundle. But with this code they will.

How do we want this to work with accident insurance? If you have both apartment AND house insurance AND an accident insurance - which combinations will become the bundle? Will that depend on the address value of the accident insurance?

src/resolvers/contracts.test.ts
```diff
@@ -405,6 +482,15 @@ describe('Query.contracts', () => {
     expect(result).toMatchSnapshot()
   })
 
+  it('SwedishAccident', async () => {
```
robinandeer, line 485: What is this test asserting?

sampettersson (approved):

PR #1329, opened 2021-08-31
Type of contract gradient

sampettersson:
## What?
- Rename QuoteBundleAppConfigurationGradientOption to TypeOfContractGradientOption and add gradient option as a field to Contract.

## Why?
- For the apps to know which gradientOption to show 

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [ ] Codescouted


hedvigoscar (commented):
While I'm not opposed to this solution, perhaps one step more generic could be to actually deliver the color components of the gradients? Then we can release entirely new gradients for new contract types 😏 

Something like this, perhaps?

```graphqls
type Gradient {
    dark: GradientColors!
    light: GradientColors!
}
type GradientColors {
    start: String!
    end: String!
}
```

Counter-pitch would be that it's a little bothersome to do. What do you think? 🤔

hedvigoscar (approved):
We explored it and it was bothersome to implement. This strikes a good balance.

PR #1330, opened 2021-09-01
Suggestion / Bundle SE contracts based on address

robinandeer:
## What?

- Add test for member with two bundles: apartment + accident AND house.
- Update fixture data to use different addresses for apartment and house agreements.
- Update snapshots to match new fixture data.

## Why?

- See discussion in #1327 

## Optional checklist
- [x] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


src/resolvers/contracts.ts
```diff
@@ -169,6 +167,10 @@ const getPotentialCrossSells = (
   return []
 }
 
+const getAddressKey = (address: AddressDto): string => {
+  return `${address.street} ${address.postalCode} ${address.country}`
```
bystam, line 171: Do you think we should fingerprint more aggressively? For instance strip away whitespace and lowercase everything or similar?

robinandeer: Yeah sure, please keep working on it! I trust you when it comes to these things :)

sampettersson (approved):

PR #1332, opened 2021-09-02
Feature/cross sell embark

cpiehl1:
# Jira Issue: [MX-260]

## What?
- Move embark schema from embark to giraffe
- `CrossSellEmbark` now has `EmbarkStory` instead of id
- Added accident Sweden to `potentialCrossSells`

## Why?
- We need `EmbarkStory` in `CrossSellEmbark`, but `EmbarkStory` lies in embark's schema so to avoid stitching we move the entire schema to giraffe

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [x] Codescouted

[MX-260]: https://hedvig.atlassian.net/browse/MX-260


src/schema.graphqls
```diff
@@ -797,7 +802,8 @@ type CrossSellChat {
 }
 
 type CrossSellEmbark {
-  embarkStoryId: ID!
+  name: String!
+  embarkStory(locale: Locale!): EmbarkStory
```
cpiehl1, line 806: Here I added name because it is set while resolving `potentialCrossSells` and I also need it in the resolver for embarkStory. Perhaps there is a cleaner way to solve this.

src/resolvers/cross-schema/index.ts
```diff
@@ -422,7 +421,7 @@ const quotesExtension: CrossSchemaExtension = {
 }
 
 const embarkExtension: CrossSchemaExtension = {
-  dependencies: [SchemaIdentifier.CONTENT_SERVICE, SchemaIdentifier.EMBARK],
+  dependencies: [SchemaIdentifier.CONTENT_SERVICE],
```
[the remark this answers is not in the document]
hedvigoscar: The easiest way to test this is probably to run `giraffe` with `content-service` running as well, and attempting to resolve `EmbarkPreviousInsuranceProviderActionData.insuranceProviders`. 🙂

src/schema.graphqls
cpiehl1: Wow, thats very clean!

src/resolvers/cross-schema/index.ts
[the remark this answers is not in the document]
cpiehl1: Seems to be working!

src/schema.graphqls
hedvigoscar: It would have been nicer if TypeScript played ball, but it works decently well 🤷

src/schema.graphqls
```diff
@@ -797,7 +802,7 @@ type CrossSellChat {
 }
 
 type CrossSellEmbark {
-  embarkStoryId: ID!
+  embarkStory(locale: Locale!): EmbarkStory
```
hedvigoscar, line 805:
```suggestion
  embarkStory(locale: Locale!): EmbarkStory!
```

I'd make this required - if we know enough to say that the cross-sell is provided by embark, then I think we should be able to guarantee that the story is there as well 😉

cpiehl1:
@hedvigoscar Note that there is a small difference in the queries, compared to the embark schema, `locale: String!` is now `locale: Locale!`, Sam asked me to make this change.

src/schema.graphqls
cpiehl1: True 😄

src/schema.graphqls
cpiehl1: Yea the red lines hurt my soul 😢

src/resolvers/contracts.ts
```diff
@@ -30,6 +32,22 @@ const ADDRESS_CHANGE_STORIES_BY_MARKET: Record<string, string> = {
...
+
+export const getCrossSellEmbarkStory: CrossSellEmbarkResolvers['embarkStory'] = async (
+  parent,
+  { locale },
+  { upstream },
+) => {
+  const parentAsPrivate = parent as CrossSellEmbarkPrivate
+  if (!parentAsPrivate._name) {
+    return undefined
+  }
```
hedvigoscar, line 47: This return should now be illegal, given the new type requirement on `CrossSellEmbark.embarkStory` 😉

src/schema.graphqls
cpiehl1: When changing to `as unknown as CrossSellEmbarkPrivate` TS doesn't complain about types not sufficiently overlapping. Would this ever be a good idea?

cpiehl1: This might stop my build from failing

sampettersson (approved):
👌🏻

PR #1333, opened 2021-09-07
Quote bundle variations

sampettersson:
# Jira Issue: [APP-766]

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [ ] Codescouted

[APP-766]: https://hedvig.atlassian.net/browse/APP-766


src/resolvers/cross-schema/quoteBundleVariations.ts
```diff
@@ -0,0 +1,152 @@
...
+    }
+  `,
+  resolvers: (schemas) => ({
+    QuoteBundleVariant: {
+      tag: {
+        resolve: (variant: QuoteBundleVariant, args: { locale: string }) => {
+          const textKeyMap = require(`${__dirname}/../../translations/${args.locale}.json`)
+
+          /// assume that a variant with more than one quote should have a popular tag
+          if (variant.bundle.quotes.length > 1) {
```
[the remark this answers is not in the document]
sampettersson: Yeah, I was thinking the same we can just modify this logic when we add NO & DK to the mix 😄

src/resolvers/cross-schema/quoteBundleVariations.ts
```diff
@@ -0,0 +1,152 @@
...
+
+interface QuoteBundle {
+  quotes: Quote[]
+}
+
+interface QuoteBundleVariant {
+  bundle: QuoteBundle
+}
+
+const SEHomeOrHouseAndAccidentBundleTypes = [
```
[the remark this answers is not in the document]
sampettersson: Fixed!

styrelius (commented):
Looks good! I don't know how to test this out locally though 🤔

src/resolvers/cross-schema/quoteBundleVariations.ts
```diff
@@ -0,0 +1,152 @@
...
+                ),
+              },
+            }
+
+            return [
+              homeAndAccidentVariant.bundle.quotes.length == 2
+                ? homeAndAccidentVariant
+                : null,
+                homeVariant,
+            ].filter((i) => i !== null)
```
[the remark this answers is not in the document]
sampettersson: @styrelius ahhh, that is actually way cleaner 😄

robinandeer:
@sampettersson cool addition! One thing we are wondering about is how we get the bundle cost for each variation. Is that something we can request per "possible variation" with this solution? Including discounts etc.

We can discuss in person if that makes more sense!

sampettersson:
@robinandeer so the thinking on that part was that each variation just returns the type `QuoteBundle` which means that you can just for example fetch price like this:

```GraphQL
query {
  quoteBundle(
    input: {
      ids: [
        "55aad267-3e46-4af1-8262-6fedbc3cf9e7"
        "8ab392b5-7f68-4520-90e3-1cea49c7b9e4"
      ]
    }
  ) {
    possibleVariations {
      tag(locale: sv_SE)
      bundle {
        cost {
          monthlyGross
        }
      }
    }
  }
}
```

src/resolvers/cross-schema/quoteBundleOrderedQuotesExtension.ts
```diff
@@ -43,13 +43,32 @@ const AddFieldToDelegatedRequest = (
 }
 
 interface Quote {
-  typeOfContract: String
+  typeOfContract: string
```
robinandeer, line 46: 🧹

src/resolvers/cross-schema/quoteBundleOrderedQuotesExtension.ts
```diff
@@ -43,13 +43,32 @@ const AddFieldToDelegatedRequest = (
...
 }
 
 interface QuoteBundle {
   quotes: Quote[]
 }
 
+const PrioritisedHomeTypeOfContracts = [
```
robinandeer, line 53: Perhaps we should type this as `Array<TypeOfContract>`?

src/resolvers/cross-schema/quoteBundleOrderedQuotesExtension.ts
```diff
@@ -43,13 +43,32 @@ const AddFieldToDelegatedRequest = (
...
+  'NO_HOME_CONTENT_RENT',
+  'NO_HOME_CONTENT_YOUTH_OWN',
+  'NO_HOME_CONTENT_YOUTH_RENT',
+  'DK_HOME_CONTENT_OWN',
+  'DK_HOME_CONTENT_RENT',
+  'DK_HOME_CONTENT_STUDENT_OWN',
+  'DK_HOME_CONTENT_STUDENT_RENT'
+]
+
+const isPrioritisedHomeQuote = (typeOfContract: string): boolean =>
```
robinandeer, line 69: ...and type this as `(typeOfContract: TypeOfContract): boolean`?

src/resolvers/cross-schema/quoteBundleVariations.ts
```diff
@@ -0,0 +1,169 @@
+import { gql } from 'apollo-server-koa'
+import { String } from 'aws-sdk/clients/cloudhsm'
+import { SchemaIdentifier, CrossSchemaExtension } from './index'
+
+interface Quote {
+  id: string
+  typeOfContract: String
```
robinandeer, line 7:
```suggestion
  typeOfContract: string
```

🧹 or type it explicitly as `TypeOfContract`?

src/resolvers/cross-schema/quoteBundleVariations.ts
```diff
@@ -0,0 +1,169 @@
...
+interface QuoteBundle {
+  quotes: Quote[]
+}
+
+interface QuoteBundleVariant {
+  id: string
+  bundle: QuoteBundle
+}
+
+const SEHomeAndAccidentBundleTypes = [
```
robinandeer, line 20: These could also be types explicitly as arrays of `TypeOfContract` enum values, no?

src/resolvers/cross-schema/quoteBundleVariations.ts
```diff
@@ -0,0 +1,169 @@
...
+]
+
+const quotesToID = (quotes: Quote[]): string => 
+  quotes.reduce((prev, curr) => {
+    if (prev.length == 0) {
+      return curr.id
+    }
+    
+    return `${prev}+${curr.id}`
+  }, "")
```
robinandeer, line 45: Isn't this simpler?

```suggestion
const quotesToID = (quotes: Quote[]): string => 
  quotes.map((item) => item.id).join('+')
```

robinandeer (commented):
From what I can read I think this makes sense.

I think we should have an idea about how to resolve the "tag" field across multiple markets but I guess it's okey to start with this solution.

robinandeer (approved):
Great work! From my point it looks good now :D



We can add the bundle logic for DK/NO later as discussed

PR #1336, opened 2021-09-08
Feature/cross sell info

cpiehl1:
# Jira Issue: [MX-257]

## What?
- Add `info` to `potentialCrossSells`, which contains the following 
  - `perils`
  - `terms`
  - `highlights`
  - `faq`

## Why?
- So that the apps can show additional information about the product in the cross-selling flow

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [ ] Codescouted

[MX-257]: https://hedvig.atlassian.net/browse/MX-257


src/resolvers/contracts.ts
```diff
@@ -38,6 +39,34 @@ interface CrossSellEmbarkPrivate {
   _name: string
 }
 
+interface CrossSellInfoPrivate {
+  _contractType?: string
+}
+
+export const crossSellResolver: Resolvers['CrossSell'] = {
+  info: async (parent, { locale }, { upstream }) => {
+    const parentAsPrivate = (parent as unknown) as CrossSellInfoPrivate
```
rasmusguterstam, line 48: I don't know how consistent the rest of this project is on this point, but perhaps go for 1 resolver = 1 file?

rasmusguterstam (approved):
Just cross-sell me an accident already 💸

src/resolvers/contracts.ts
cpiehl1: Yes, this was a bit weird since the rest of the cross-sell is resolved on `contractBundle`, but I agree that moving these to a separate file makes sense.

PR #1337, opened 2021-09-09
Add promisePerils as replacement for perils

joacimastrom:
# Jira Issue: [IPC-221]

## What?
- Create new query for perils served from promise-cms

## Why?
- Urls for icons change from relative to absolute when switching from content-service to promise-cms. To stay backwards compatible with older app-clients, we create a new query to use going forward with clients that support absolute paths. 

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [ ] Codescouted

[IPC-221]: https://hedvig.atlassian.net/browse/IPC-221


src/schema.graphqls
```diff
@@ -617,6 +617,18 @@ type Member {
   features: [Feature!]!
 }
 
+type InsuranceTerm {
```
cpiehl1, line 620: @hedvigoscar @sampettersson I will use this in the cross-selling flow but I'm not sure if maybe I should call it `InsuranceTermV2` since there is currently a `type InsuranceTerm` in the schema in content-service. However, my understanding is that as long as the types look the same there won't be any problems with the stitching.

joacimastrom: Since they are identical, I think it's fine just overriding it like this :) That's what we did when we moved perilsV2 from content-service as well

src/api/upstreams/promiseCms.ts
```diff
@@ -38,31 +59,24 @@ type Media = {
 }
 
 // Our locales are not available among the standard i18n that Strapi exposes. We could override Strapi to match ours if we want to
```
tobiasbexelius, line 61: Orphaned comment

tobiasbexelius, line 61: Orphaned comment

src/generated/graphql.ts
```diff
@@ -3822,7 +3846,7 @@ export type QueryResolvers<ContextType = Context, ParentType extends ResolversPa
   selfChangeEligibility?: Resolver<ResolversTypes['SelfChangeEligibility'], ParentType, ContextType>;
   availableLocales?: Resolver<Array<ResolversTypes['Locale']>, ParentType, ContextType>;
   claims?: Resolver<Array<ResolversTypes['Claim']>, ParentType, ContextType>;
-  perilsV2?: Resolver<Array<ResolversTypes['PerilV2']>, ParentType, ContextType, RequireFields<QueryPerilsV2Args, 'contractType' | 'locale'>>;
```
tobiasbexelius, line 3825: Who's calling this? Shouldn't it be kept around until clients have switched to the new query?

tobiasbexelius (approved):
Great! Just make sure that the removed query is not being called and remove that comment before you merge

src/generated/graphql.ts
joacimastrom: perilsV2 was the first version of `promisePerils` and is not yet in use. All clients still use `perils`, but have been instructed to use promisePerils going forward :)

PR #1340, opened 2021-09-13
Add Swedish Accident to contractDetailsFragments

cpiehl1:
# Jira Issue: []

## What?
- Adding Swedish Accident to `contractDetailsFragments`

## Why?
- So that this can be resolved by clients

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [ ] Codescouted


cpiehl1:
@hedvigoscar This will hopefully fix your query. I am wondering if it should also be added to `upcomingAgreementChangeFragment`? Also, I noticed that `NorwegianTravel`, `DanishAccident` and `DanishTravel` are also missing from these fragments, should they be added too?

hedvigoscar:
I think that would be wise, yes! 🙂

hedvigoscar (approved):

PR #1342, opened 2021-09-13
Get insuranceTerms and t&cs from promise-cms instead of content-service

joacimastrom:
# Jira Issue: [IPC-242]

## Dependant PRs:
[api-gateway](https://github.com/HedvigInsurance/api-gateway/pull/58)
[prod-env](https://github.com/HedvigInsurance/prod-env/pull/194)

## What?
- Switch insuranceTerms and termsAndConditions from content-service to promise-cms

## Why?
- Routes will fail when we eventually kill content-service

## Optional checklist
- [ ] Unit tests written
- [X] Tested locally
- [ ] Codescouted

[IPC-242]: https://hedvig.atlassian.net/browse/IPC-242


src/resolvers/cross-schema/index.ts
```diff
@@ -199,41 +199,25 @@ const contractExtension: CrossSchemaExtension = {
...
-              contractType: contract.typeOfContract,
-              locale: args.locale,
-            },
-            context,
-            info,
-          })
-        },
+          { upstream },
+        ) =>
+          upstream.promiseCms.getTermsAndConditions({
```
cpiehl1, line 204: Do we need to resolve this field here now. Can't we just resolve it directly in a resolver for `contract` instead of extending the type?

src/api/upstreams/promiseCms.ts
```diff
@@ -13,6 +13,10 @@ export interface PromiseCmsClient {
     locale: string
     date?: string
   }): Promise<ContractInfo>
+  getTermsAndConditions(args: {
+    contractType: string
+    locale: string
```
cpiehl1, line 18: Do we want `date?` as well?

src/resolvers/cross-schema/index.ts
```diff
@@ -199,41 +199,25 @@ const contractExtension: CrossSchemaExtension = {
...
-            schema: schemas(SchemaIdentifier.CONTENT_SERVICE),
-            operation: 'query',
-            fieldName: 'insuranceTerms',
-            args: {
-              contractType: contract.typeOfContract,
-              locale: args.locale,
-            },
-            context,
-            info,
+          const { terms } = await upstream.promiseCms.getContractInfo({
```
cpiehl1, line 216: Same here, maybe we don't need this extension anymore.

src/resolvers/cross-schema/index.ts
```diff
@@ -325,41 +309,25 @@ const quotesExtension: CrossSchemaExtension = {
...
-              contractType: quote.typeOfContract,
-              locale: args.locale,
-            },
-            context,
-            info,
-          })
-        },
+          { upstream },
+        ) =>
+          upstream.promiseCms.getTermsAndConditions({
```
cpiehl1, line 314: Same here.

src/resolvers/cross-schema/index.ts
```diff
@@ -422,41 +390,25 @@ const quotesExtension: CrossSchemaExtension = {
...
```
cpiehl1, line 395: And here

src/resolvers/cross-schema/index.ts
```diff
@@ -422,41 +390,25 @@ const quotesExtension: CrossSchemaExtension = {
...
-            schema: schemas(SchemaIdentifier.CONTENT_SERVICE),
-            operation: 'query',
-            fieldName: 'insuranceTerms',
-            args: {
-              contractType: quote.typeOfContract,
-              locale: args.locale,
-            },
-            context,
-            info,
+          const { terms } = await upstream.promiseCms.getContractInfo({
```
cpiehl1, line 407: And here

cpiehl1 (commented):
Looks good! I am just wondering if we can't reduce some of the cross-schema extensions now.

src/api/upstreams/promiseCms.ts
joacimastrom: Date wasn't in the older implementation so I left that out :shrug:

src/resolvers/cross-schema/index.ts
joacimastrom: Yeah you're right, that's a lot better!

joacimastrom: Hmm actually no. Since the `locale` parameter is on the extended queries and not on the `contracts`-query itself :/ It is absolutely doable and better, but it would also be a breaking change 😬

hedvigoscar: You can probably define a resolver for `Contract.termsAndConditions` and have that take `Locale` as an argument - that would be entirely schema-compatible 🙂

joacimastrom: Could we huddle this real quick after standups? :)

cpiehl1: I am available :)

src/resolvers/contractPerils.ts
```diff
@@ -1,21 +1,29 @@
...
   _,
-  args: PerilsArgs,
+  args: QueryContractPerilsArgs,
   { upstream }: { upstream: Upstream },
 ): Promise<PerilV2[]> => {
   const { perils } = await upstream.promiseCms.getContractInfo(args)
   return perils
 }
+
+export const contractPerilExtension: ContractResolvers['contractPerils'] = async (
```
cpiehl1, line 19: Maybe this should be called `contractPerilResolver` now?

src/resolvers/contractPerils.ts
```diff
@@ -1,21 +1,29 @@
...
+  parent,
+  args: ContractContractPerilsArgs,
+  { upstream }: { upstream: Upstream },
```
cpiehl1, line 22:
```suggestion
  { upstream },
```

And then not import `Upstream`?

cpiehl1 (approved):
Looks good 👍

src/resolvers/contractPerils.ts
joacimastrom: Following the pattern of all other resolvers, their naming is just what they resolve to :)

sampettersson (approved):

PR #1345, opened 2021-09-14
MX-274 move cross-selling logic to underwriter

cpiehl1:
# Jira Issue: [MX-274]

## What?
- Moving cross-selling logic to underwriter

## Why?
-

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [ ] Codescouted

[MX-274]: https://hedvig.atlassian.net/browse/MX-274


src/resolvers/contracts.ts
```diff
@@ -274,10 +148,7 @@ export const bundleContracts = (
           addressChangeAngelStoryId &&
           `${addressChangeAngelStoryId}?contractBundleId=${bundleId}`,
       },
-      potentialCrossSells:
-        Object.values(swedishBundle).length < 2
-          ? getPotentialCrossSells(contracts, strings)
-          : [], // We only cross-sell if there is a single bundle
+      potentialCrossSells: [],
```
rasmusguterstam, line 151: What's this? 😮

rasmusguterstam (approved):
Don't know how it looks in underwriter, but this looks very nice! 🔥

src/resolvers/contracts.ts
cpiehl1: Our codegen library does not allow returning partial types on resolvers, so I will get a typescript error by omitting that. There is a plugin for returning partial types which could be used instead but it felt like overkill for this implementation.

PR #1352, opened 2021-09-17
Download new translations

styrelius:
# ~Jira Issue: []~

## What?
- New translations and text keys downloaded.

## Why?
- Mainly since the Purchase Journey team would like to be able to use the `CONTRACT_DISPLAY_NAME_SE_ACCIDENT` for the `displayName` property of Swedish accident insurance.

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted

## Note
I don't yet know how to run Giraffe locally (I tried and failed and would be very grateful if I could get some help with that) and I don't know if there's some risk with just adding new translations without checking them out locally but I assume it should be fine?


styrelius:
Never mind, we already has those translations apparently, if anyone else wants the new translations downloaded here I guess you can re-open this 😄

PR #1364, opened 2021-09-29
Add onboarding campaign mutations

tobiasbexelius:
# Jira Issue: [MX-288]

## What?
- Mutations to add and remove campaigns
- Uses the header X-Onboarding-Session instead of token

## Why?
- So that we can support campaigns in the memberless onboarding

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [x] Codescouted

[MX-288]: https://hedvig.atlassian.net/browse/MX-288


tobiasbexelius:
See https://github.com/HedvigInsurance/giraffe/pull/1383 instead

PR #1369, opened 2021-09-29
Return current co-insured when cross-selling

cpiehl1:
# Jira Issue: []

## What?
- Return current number of co-insured on the base contract when cross-selling

## Why?
- So that this value can be default in the embark flow

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


src/resolvers/crossSell.ts
```diff
@@ -69,7 +75,9 @@ const getCrossSell = (type: string, strings: LocalizedStrings): CrossSell => {
       callToAction: strings('UPSELL_NOTIFICATION_ACCIDENT_CTA'),
       action: ({
         __typename: 'CrossSellEmbark',
-        _name: `accident-cross-sell-${type.slice(0, 2)}`,
+        _name: `accident-cross-sell-${type.slice(0, 2)}${
+          currentCoInsured ? `?coInsured=${currentCoInsured}` : ''
```
cpiehl1, line 79: Is `coInsured` the correct key? Not sure how this is read inside the embark flow. @hugokallstrom

hugokallstrom: Looking at embark and the angel story, we already have a key called `householdSize`. This represents the input from the user. It would be really smooth if we could map this value to that key - so instead of passing `coInsured` here we pass `householdSize`. Does that make sense? Sorry for the hassle!

hugokallstrom (changes requested):

src/resolvers/crossSell.ts
cpiehl1: Yes, this should be fine. How is it going to work for DK and NO thought? When creating quotes, `householdSize` is used in SE and `coInsured` in DK and NO. Is the embark flow `householdSize` regardless of market?

hedvigoscar: We're using `householdSize` in the embark story as it makes more sense for the user. Later on, when submitting the API call, we calculate `numberCoinsured` instead by `householdSize - 1`

cpiehl1: Alright, sounds good! 😄

vonElfvin (approved):
Izi

hugokallstrom (approved):

PR #1375, opened 2021-10-05
Fix/moving flow input validation

cpiehl1:
# Jira Issue: []

## What?
- Better error logging when moving flow input is invalid

## Why?
-

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


vonElfvin (approved):

src/resolvers/addressChange.ts
```diff
@@ -260,3 +260,16 @@ const transformResult = (
...
+  (fieldName: string, value?: string): string
+  (fieldName: string, value?: boolean): boolean
+}
+
+const requiredField: IRequired = (fieldName: string, value?: any) => {
+  if (value == null) {
+    throw Error(`Field: ${fieldName} is required but was not provided`)
+  }
+  return value
+}
```
bystam, line 275:
```suggestion
function requiredField<T>(fieldName: string, value: T | null | undefined) {
  if (!value) {
    throw Error(`Field: ${fieldName} is required but was not provided`)
  }
  return value
}
```

This simpler perhaps? :)

bystam (approved):
Looks good! Had one cosmetic suggestion that I think works :P

src/resolvers/addressChange.ts
cpiehl1: Yes! I wanted to do something like this but didn't get it to work 😄

cpiehl1: Maybe because I was using an arrow function?

cpiehl1: But I have to use `value == null`, otherwise it fails when `value = false`

PR #1377, opened 2021-10-06
Add about section to cross sell info

cpiehl1:
# Jira Issue: []

## What?
- Adding an about section to `CrossSellInfo`

## Why?
-

## Optional checklist
- [x] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


src/resolvers/crossSell.ts
```diff
@@ -29,6 +29,9 @@ export const crossSellResolver: Resolvers['CrossSell'] = {
 
     return {
       displayName: strings(`CONTRACT_DISPLAY_NAME_${parent.contractType}`),
+      aboutSection: strings(
+        `CROSS_SELL_ABOUT_${parent.contractType.split('_')[1]}`,
```
cpiehl1, line 33: I assumed this should come from lokalise

cpiehl1: My reasoning for choosing lokalise was that it is just a static text, for which PCMS seems a bit overkill

cpiehl1: I also think that PCMS might become a bit cluttered if we start including this type of content

src/api/upstreams/promiseCms.ts
```diff
@@ -43,6 +37,7 @@ interface InsurableLimitCollection {
 
 interface ContractInfoDto {
   contract_type: string
+  about_section?: string
```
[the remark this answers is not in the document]
cpiehl1: Good catch 😅

PR #1380, opened 2021-10-11
MX-316: Added new GraphQL queries and mappings which were migrated from content-service.

maksym-krupenko:
# Jira Issue: [MX-316]

## What?
- Added new GraphQL queries and mappings which were migrated from content-service.

## Why?
- These changes are related to global migration of GraphQL code from services to Giraffe. Details could be found here: https://www.notion.so/hedviginsurance/Exposing-APIs-to-Giraffe-6ac290ac76934ab588e87da45870d104

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted

[MX-316]: https://hedvig.atlassian.net/browse/MX-316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/api/upstreams/contentService.ts
```diff
@@ -0,0 +1,54 @@
...
+  CommonClaim,
+  InsuranceProvider,
+  News,
+  PerilV2,
+  ReferralTerm,
+  Welcome
+} from "../../generated/graphql";
+
+export interface ContentServiceClient {
+  commonClaims(locale: string): Promise<CommonClaim[]>
```
cpiehl1, line 13: How does this work when `CommonClaim` has a union type inside? Are we able to parse it from JSON?

cpiehl1 (approved):
Very nice! 💯 Just have one question

src/api/upstream.ts
```diff
@@ -40,7 +44,8 @@ export interface Upstream {
   underwriter: UnderwriterClient
   claimService: ClaimServiceClient
   promiseCms: PromiseCmsClient
-  embark: EmbarkClient
+  embark: EmbarkClient,
+  contentServiceClient: ContentServiceClient
```
bystam, line 48: The other properties omit the `Client` part in its name as it becomes quite obvious at the call site anyway:
```typescript
const res = await upstream.contentService.commonClaims(...)
```

PR #1385, opened 2021-10-13
Support multi-quote display names for bundles

robinandeer:
# Jira Issue: [GRW-447]

## What?

- Add support for defining multi-quote bundle display names in Lokalise

## Why?

- We want customised display names for each type of bundle
- This solution will fallback to existing behaviour if the bundle display name is not found in Lokalise

## Optional checklist

- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


sampettersson (approved):
Makes sense!

PR #1387, opened 2021-10-13
Order possible variations quotes

robinandeer:
# Jira Issue: [GRW-447]

## What?

- Order quotes under `possibleVariations` the same as on the root level

## Why?

- We should be able to expect the same ordering for quotes
- We need to trust the order to setup bundle specific display names in Lokalise #1385 

## Optional checklist

- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


sampettersson (approved):
Ohh, sweet!

src/resolvers/cross-schema/index.ts
```diff
@@ -67,8 +67,8 @@ export const getCrossSchemaExtensions = (
     createContractPills(),
     createQuoteFaqsExtension(),
     createQuoteBundleAppConfigurationExtension(),
-    createQuoteBundleDisplayNameExtension(),
     createQuoteBundleOrderedQuotesExtension(),
+    createQuoteBundleDisplayNameExtension(),
```
robinandeer, line 71: @sampettersson does this order make a difference? We want to make sure that ordering happens before setting the display name for bundles 🙃

robinandeer:
This was resolved in #1392

PR #1417, opened 2021-11-05
make nullable first and last name in bundle quotes

meletios:
# Jira Issue: []

## What?
- Connected to https://github.com/HedvigInsurance/underwriter/pull/368

## Why?
- To enable web to call the bundleQuote endpoint without a first and last name for our partners

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


src/generated/graphql.ts
```diff
@@ -210,8 +210,8 @@ export type BundledQuote = {
   id: Scalars['ID'];
   currentInsurer?: Maybe<CurrentInsurer>;
   price: MonetaryAmountV2;
-  firstName: Scalars['String'];
-  lastName: Scalars['String'];
```
cpiehl1, line 214: Hmm, did you change this file manually? How come there is no change to the schema?

mehdielket (approved):
Looks good!

robinandeer (commented):
I know we have a discussion in Slack about this but should we really break backwards compatibility if we can instead use different fields where firstname/lastname are optional?

sampettersson:
I wonder what actually happens in prod, if we in practice rarely send null, then it should work even though we don't update the clients, right? So we could just update Codegen behind the scenes, adapt the code and we good?

robinandeer:
I would still prefer making non-breaking changes and deprecating these fields.

My gut feeling is that it wouldn't cause big issues. However, it is still an assumption and web can simply push out a hot fix if we are wrong -- so if apps are okey with it? ;)

hedvigoscar:
I agree with @robinandeer in the general case - breaking changes should be limited only to the very necessary, and even then after a longer deprecation period where all clients have had ample time to migrate to alternative APIs.

In this particular case I'm not opposed to the change, from the perspective of the android-platform, as we never query these fields.

mehdielket:
Same here. I agree with you @robinandeer in principle, but I think in this specific case this will not affect us either, so if we can avoid deprecation all together than that's better

robinandeer:
Okey, it seems I'm outnumbered 😉

Let's go with this solution and test that our assumptions hold up in staging!

guilhermespopolin (approved):
Let's test it on stage

robinandeer:
Why can't we test in a review app? @guilhermespopolin

PR #1421, opened 2021-11-16
APP-1074: Claims status card view model

StylianosGakis:
# Jira Issue: [APP-1074](https://hedvig.atlassian.net/jira/software/c/projects/APP/boards/6?modal=detail&selectedIssue=APP-1074&quickFilter=51)

## What?
- Add a new query type to get the Claim status card's view model directly from Giraffe
- In the process, I also moved the "claims" related stuff to it's own categorized folder under resolvers/claims from resolvers/uncategorized

## Why?
- Enable changing the UI without necessarily releasing new App builds.
- Remove duplicated logic from clients

## Optional checklist
- [X] Unit tests written
- [X] Tested locally
- [X] Codescouted

This is my first time doing TS ever, so while reviewing, be nit-picky about anything that I've written in a non-TS way too.


src/resolvers/claims/status-card/statusCard.ts
```diff
@@ -0,0 +1,248 @@
...
+      args: { locale: string },
+      { upstream, strings },
+    ): Promise<ClaimStatusCard[]> => {
+      const claims = (await upstream.claimService.getMemberClaims()).sort(
+        (a, b) =>
+          new Date(a.registrationDate).getTime() -
+          new Date(b.registrationDate).getTime(),
+      )
+      const localizedStrings = strings.withLocale(args.locale)
+
```
hedvigoscar, line 29: I believe we should do some filtering here - for example, claims that have been closed for `n` days should not be returned here. Check in with Hiro for the exact business rule 🙂

src/resolvers/claims/status-card/statusCard.ts
```diff
@@ -0,0 +1,248 @@
...
+      return Promise.all(
+        claims.map((c) =>
+          transformClaimStatusCard(c, args.locale, localizedStrings),
+        ),
+      )
+    },
+  },
+}
+
+// TODO This doesn't even need to be async right? Any reason we'd keep it async if it doesn't await for anything?
```
hedvigoscar, line 39:
```
// TODO This doesn't even need to be async right? Any reason we'd keep it async if it doesn't await for anything?
```

That's right. You can simply remove the `async` keyword, and the `Promise` from the return type signature.

Additionally, you'll want to remove the `Promise.all`-part on line 30. 🙂

src/resolvers/claims/status-card/statusCard.ts
```diff
@@ -0,0 +1,248 @@
...
+    progressSegments: transformProgressSegments(claim, strings),
+  } as ClaimStatusCard
+}
+
+const transformClaimStatusPills = (
+  claim: ClaimDto,
+  locale: string,
+  strings: LocalizedStrings,
+): ClaimStatusCardPill[] => {
+  const pills: ClaimStatusCardPill[] = claimStatusPillsFromClaimDto(
```
hedvigoscar, line 59:
```suggestion
  const pills = claimStatusPillsFromClaimDto(
```

This type information should probably be implicit from the call 🙂

src/resolvers/claims/status-card/statusCard.ts
```diff
@@ -0,0 +1,248 @@
...
+  locale: string,
+  strings: LocalizedStrings,
+): Promise<ClaimStatusCard> => {
+  return {
+    id: claim.id,
+    pills: transformClaimStatusPills(claim, locale, strings),
+    title: transformClaimTitle(claim, strings),
+    subtitle: transformClaimSubTitle(claim),
+    progressSegments: transformProgressSegments(claim, strings),
+  } as ClaimStatusCard
```
hedvigoscar, line 51:
```suggestion
  }
```

Is this cast really necessary? I fear that we might be 'tricking' TypeScript into giving us worse type checks by making the cast.

src/resolvers/claims/status-card/statusCard.ts
```diff
@@ -0,0 +1,248 @@
...
+}
```
hedvigoscar, line 52:
```suggestion
): Promise<ClaimStatusCard> => ({
    id: claim.id,
    pills: transformClaimStatusPills(claim, locale, strings),
    title: transformClaimTitle(claim, strings),
    subtitle: transformClaimSubTitle(claim),
    progressSegments: transformProgressSegments(claim, strings),
  })
```

Single-statement arrow functions in JS/TS need no curlies, in the same way as such functions work in Kotlin 🙂

src/resolvers/claims/status-card/statusCard.ts
```diff
@@ -0,0 +1,248 @@
...
+        {
+          text: strings('home.claim_card.pill.reopened'),
+          type: ClaimStatusCardPillType.Reopened,
+        } as ClaimStatusCardPill,
+        {
+          text: strings('home.claim_card.pill.claim'),
+          type: ClaimStatusCardPillType.Open,
+        } as ClaimStatusCardPill,
+      ]
+    // TODO should this exist in the case that we add another claim status for backwards compatibility reasons?
```
hedvigoscar, line 150: Unknown outcomes can probably return nothing for the time being, as we can trivially add support for them later. 🙂

hedvigoscar (commented):
A few things to look at 🙂

src/resolvers/claims/status-card/statusCard.ts
StylianosGakis: We've had this discussion before, wasn't sure if it were to be done in this layer. But you're right, we don't want to change the service API as you've said [here](https://hedviginsurance.slack.com/archives/C016KUN61U3/p1636634598109300?thread_ts=1636626553.105100&cid=C016KUN61U3) as well, so I'll make it happen here instead. I'll ask Hiro about the time specifics

src/resolvers/claims/status-card/statusCard.ts
StylianosGakis: I thought this was like *the* way to create an instance of the `(export) type ClaimStatusCard` because none of all the other syntax I tried, like: `new ClaimStatusCard()` worked 😂
I guess you just wrap it in {} and let the IDE check it's type by comparing it with the function's return type. Interesting for sure 👀

src/resolvers/claims/status-card/statusCard.ts
hedvigoscar: This is indeed the appropriate layer, that's a large part of why we're building this thing, actually 😉

src/resolvers/claims/status-card/statusCard.ts
hedvigoscar: Yes! The concept is called _structural typing_ (as opposed to _nominal typing_). Quite neat! More reading here: https://www.typescriptlang.org/docs/handbook/type-compatibility.html

src/resolvers/claims/status-card/statusCard.ts
```diff
@@ -0,0 +1,234 @@
...
+        case ClaimOutcomeDto.PAID: {
+          const pillData: ClaimStatusCardPill[] = []
+          pillData.push({
+            text: strings('claim.decision.paid'),
+            type: ClaimStatusCardPillType.Closed,
+          })
+          const payout = claim.payout
+          if (payout != null) {
+            // TODO How to do currency conversion properly? format() needs a number|bigint but we get a string from the
+            //  service. We do this exact thing in campaignDisplayValue.ts:82 but its type is "any" so it doesn't care?
```
hedvigoscar, line 101: Can't speak for this unfortunately 😞

hedvigoscar (approved):
LGTM with caveat - check that with MX, as discussed in person

hedvigoscar:
Also, there's a test failure, you'll want to check that out 😉

src/resolvers/claims/status-card/statusCard.ts
StylianosGakis: Yeap, I was hoping someone from the MX team can take a look at the PR as well, maybe they'll have more insight on this

src/resolvers/claims/status-card/statusCard.ts
```diff
@@ -0,0 +1,234 @@
...
+import { differenceInDays } from 'date-fns'
+
+export const ClaimsStatusCardsSubset: ResolverSubset = {
+  Query: {
+    claims_statusCards: async (
+      _parent,
+      args: { locale: string },
+      { upstream, strings },
+    ): Promise<ClaimStatusCard[]> => {
+      const localizedStrings = strings.withLocale(args.locale)
```
cpiehl1, line 24: The `strings` you get in the context are already localized using the `accept-language` header, so maybe you don't this line? Unless there are cases where these might diverge.

cpiehl1, line 24: The `strings` you get in the context are already localized using the `accept-language` header, so maybe you don't this line? Unless there are cases where these might diverge.

src/resolvers/claims/status-card/statusCard.ts
```diff
@@ -0,0 +1,234 @@
...
+  ClaimStatusDto,
+} from '../../../api/upstreams/claimService'
+import { LocalizedStrings } from '../../../translations/LocalizedStrings'
+import { differenceInDays } from 'date-fns'
+
+export const ClaimsStatusCardsSubset: ResolverSubset = {
+  Query: {
+    claims_statusCards: async (
+      _parent,
+      args: { locale: string },
```
cpiehl1, line 21:
```suggestion
      { locale },
```
So you don't need to write `args.locale` everywhere 😄

src/resolvers/claims/status-card/statusCard.ts
cpiehl1: Maybe `Number(payout.amount)`? Then you can also check with `isNaN` and throw an error

cpiehl1 (approved):
🚀

PR #1425, opened 2021-11-18
Replace OnboardingSession with QuoteCart

tobiasbexelius:
# Jira Issue: [MX-392]

## What?
- Replace onboarding session with quoteCart
- There are some things left to do, but I think it's useful to get it in at this point so the branches don't diverge

## Why?
- To enable other quote creation flows to use the same API as onboarding in the future

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted

[MX-392]: https://hedvig.atlassian.net/browse/MX-392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/resolvers/quote-bundle/quoteBundle.graphqls
```diff
@@ -140,3 +140,20 @@ type QuoteBundleVariant {
     id: ID!
     bundle: QuoteBundle!
 }
+
+enum ApartmentType {
```
robinandeer, line 144: Should this be `SwedishApartmentType` for consistency?

src/resolvers/quote-cart/quoteCart.graphqls
```diff
@@ -92,89 +102,46 @@ enum OnboardingSessionSignStatus {
     FAILED
 }
 
-input CreateOnboardingSessionInput {
-    country: String!,
+input CreateOnboardingQuoteCartInput {
+    market: String!,
```
robinandeer, line 106: I guess this is not an enum since we want to stay generic?

src/resolvers/quote-cart/quoteCart.ts
```diff
@@ -0,0 +1,286 @@
...
+      const sessionDto = await upstream.underwriter.getOnboardingSession(id)
+      // delete old quotes
+      if (sessionDto.quoteIds.length !== 0) {
+        await upstream.underwriter.updateOnboardingSession(id, {
+          quoteIds: null,
+        })
+      }
+
+      for (const quoteData of input.quoteData) {
+        const result = await upstream.underwriter.createQuote({
```
robinandeer, line 54: This will handle each create quote sequentially, right? For performance reasons we might want to run them in parallel with `Promise.all`.

robinandeer (commented):
Very cool to see this come together! :D 

I reviewed what I could but didn't find anything critical that would keep us from merging

src/resolvers/quote-cart/quoteCart.graphqls
tobiasbexelius: We could make it an enum, we don't enter new markets that often

src/resolvers/quote-cart/quoteCart.ts
tobiasbexelius: Very good idea. I'm a JS noob so let me know if there is a nicer way to do it than what I just did

hugokallstrom (approved):
Nice, LGTM!

PR #1426, opened 2021-11-19
APP-1080: Create different cross sells depending on type of contract

hugokallstrom:
# Jira Issue: [APP-1080](https://hedvig.atlassian.net/jira/software/c/projects/APP/boards/6?modal=detail&selectedIssue=APP-1080&quickFilter=51)

## What?
- Create different cross sells depending on type of contract.

## Why?
- We need to be able to have different images, embark stories etc. for specific cross sells. 

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


src/resolvers/uncategorised/crossSell.ts
```diff
@@ -75,81 +76,149 @@ const getCrossSell = (
...
+): Partial<CrossSell> => {
+  return {
+    type: CrossSellType.Accident,
+    contractType: type as TypeOfContract,
+    title: title,
+    description: description,
+    imageUrl: imageUrl,
+    blurHash: blurHash,
+    callToAction: callToAction,
+    action: ({
```
hugokallstrom, line 197: Would be nice to set 
```
action: {
      __typename: 'CrossSellChat',
    }
```
here if storyName is null. But unsure about the syntax 🤔

src/resolvers/uncategorised/crossSell.ts
```diff
@@ -75,81 +76,158 @@ const getCrossSell = (
...
-      callToAction: strings('UPSELL_NOTIFICATION_ACCIDENT_CTA'),
-      action: ({
-        __typename: 'CrossSellEmbark',
-        _name: `accident-cross-sell-${market}${
-          currentCoInsured != null
-            ? `?householdSize=${currentCoInsured + 1}`
-            : ''
-        }`,
-      } as unknown) as CrossSellEmbark,
+  switch (type) {
```
hugokallstrom, line 79: the strings we get from underwriter maps 1:1 against `TypeOfContract` right? This switch does not work though, is there a neat way to solve it?

cpiehl1: `type as TypeOfContract`

src/resolvers/uncategorised/crossSell.ts
cpiehl1:
```
action: storyName ? {
   ...
} : {
   ...
}
```

cpiehl1 (approved):
Looks great!

PR #1432, opened 2021-11-24
Improve error messages when creating bundles

tobiasbexelius:
# Jira Issue: [MX-395]

## What?
- Report the type of the quote that failed to be created
- Simplify schema for quote bundles
- Delete quotes when one fails

## Why?
- 🙂 

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [x] Codescouted

[MX-395]: https://hedvig.atlassian.net/browse/MX-395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/resolvers/quote-cart/quoteCart.graphqls
```diff
@@ -114,12 +114,7 @@ enum Market {
 }
 
 input CreateQuoteBundleInput {
-    quoteData: [QuoteData!]!
-}
-
-input QuoteData {
-    type: String!
-    payload: JSON!
+    payload: [JSON!]!
```
robinandeer, line 117: Should we add some inline docs here explaining how to structure the JSON blob?

robinandeer (commented):
Cool stuff!

Just one question -- can we use the upper case contract types now? Like `{ type: "SWEDISH_APARTMENT" }` instead of `{ type: "apartment" }`.

Otherwise we need to migrate to this new version of the API first and then again for the upper case contract types.

src/resolvers/quote-cart/quoteCart.graphqls
tobiasbexelius: Will add docs! Regarding the types, I can fix that later today.

tobiasbexelius: I'm not sure what to write though, because I'm not sure what the proper source of truth is at this point. So maybe we can do that later?

robinandeer: yeah for sure, I guess when we have some swagger UI for the schema, that would be great to reference from here :)

PR #1436, opened 2021-11-29
Fetch cashback from PCMS with locale from member-service

joacimastrom:
# Jira Issue: [IPC-390]

## What?
- Get selected cashbackOptionId and locale from member-service
- Fetch selectedCashback from PCMS
- Fetch cashbackOptions from PCMS

## Why?
- Killing content-service

## Optional checklist
- [ ] Unit tests written
- [X] Tested locally
- [X] Codescouted

[IPC-390]: https://hedvig.atlassian.net/browse/IPC-390?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/resolvers/uncategorised/cashback.ts
```diff
@@ -1,33 +1,26 @@
...
+    locale: pickedLocale,
+  } = await getSelectedCashbackOptionId(token, headersAndLocale)
 
-export const cashbackInner = async (
-  token: string,
-  headers: ForwardHeaders,
-): Promise<Maybe<Cashback>> => {
-  const options = await getCashbackOptions(token, headers)
+  console.log('locale', locale)
+  console.log('pickedLocale', pickedLocale)
```
rasmusguterstam, line 21: Is this on purpose?

rasmusguterstam (approved):
🤝

src/resolvers/uncategorised/cashback.ts
joacimastrom: Very much not so... 😅

src/api/upstreams/promiseCms.ts
```diff
@@ -125,6 +132,21 @@ export const createPromiseCmsClient = (
...
   },
+  getCashbackOptions: async (locale) => {
+    const res = await client.get(`/public/cashback-options/${locale}`, {
+      noAuthorization: true,
+    })
+    return (await res.json()) || []
+  },
+  getCashbackOption: async (cashbackOptionId, locale) => {
+    const res = await client.get(
+      `/public/cashback-option/${cashbackOptionId}/${locale}`,
```
cpiehl1, line 143: `cashback-option` here but `cashback-options` above. Is this to fetch multiple vs one?

AlexSJ1990 (approved):
✅

src/api/upstreams/promiseCms.ts
cpiehl1: Maybe a more RESTy pattern would be to do `cashback-options` for multiple and `cashback-options/{id}` for one, with the locale being a request parameter. But I guess PCMS mostly uses locale as a path variable atm so maybe it is just better to follow that pattern.

joacimastrom: yeah I agree, lets change it to plural always

joacimastrom: I think all endpoints for collections should be pluralized, so I'll change it :)

PR #1444, opened 2021-11-30
Fix campaignOwner mapping

tobiasbexelius:


robinandeer (changes requested):
Hmmm, I'm pretty sure we are using this `owner` property at least when getting `redeemedCampaigns` 🤔 or is this separate from that?

That would break our production build. Can we fix this without breaking API changes?

tobiasbexelius:
@robinandeer This is only called by quoteCart

robinandeer:
Okey understood! In that case it's okey to break the API since we don't use it in prod yet

robinandeer (approved):
If this only affects quote cart API - it looks good!

PR #1448, opened 2021-12-01
Add generic quote data to quote bundle

tobiasbexelius:
# Jira Issue: [MX-398]

## What?
- 

## Why?
-

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [x] Codescouted

[MX-398]: https://hedvig.atlassian.net/browse/MX-398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/resolvers/quote-bundle/quoteBundle.graphqls
```diff
@@ -20,7 +19,8 @@ type BundledQuote {
     lastName: String
     ssn: String
     birthDate: LocalDate!
-    quoteDetails: QuoteDetails!
+    quoteDetails: QuoteDetails! @deprecated(reason: "Use data instead.")
+    data: QuoteData
```
robinandeer, line 23: Why is this an optional field?

tobiasbexelius: Fixed!

robinandeer (approved):
I think this looks good to me :)

The only information we are missing is that fields such as `street` and `city` now are optional independently of contract type. However, like was suggested by @sampettersson we probably should not query these values directly but rather create a better model for clients to consume so we can go with this.

PR #1455, opened 2021-12-07
Map quotedto instead of bundledquote

FredrikAreschoug:
## What?
- With [this pr](https://github.com/HedvigInsurance/underwriter/pull/416) `BundledQuote` will be removed. So it's time to map `QuoteDto` instead.

## Why?
- Let giraffe take care of graphQL

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [ ] Codescouted


mightyost (approved):
Looking good to me. But are neither a Giraffe or GraphQL pro!

PR #1458, opened 2021-12-07
Return quote data as a json blob

tobiasbexelius:
# Jira Issue: []

## What?
- Return data as a json blob instead of as a union of all properties

## Why?
- To be more generic

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


mehdielket (approved):
🚀

PR #1462, opened 2021-12-09
Fix: Missing field after mapping quoteDto

FredrikAreschoug:
## What?
- Fix for missing field after mapping quoteDto
- The only thing I can see missing now is `dataCollectionId` but after some investegationg that is an issue on uw

## Why?
- :bug: 

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [ ] Codescouted


robinandeer (approved):
Seems good to me!

PR #1472, opened 2021-12-14
Fix/clean up payment dtos

cpiehl1:
# Jira Issue: []

## What?
- Clean up some dtos that weren't really used
- Rename some stuff in the schema
- Return `paymentTokenId` in `ActionRequired`

## Why?
- `paymentTokenId` is needed in the 3DS mutations and when logged in we can't get it from the quote cart

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [x] Codescouted


maksym-krupenko (approved):
Consider please change: paymentConnectionId -> paymentTokenId

PR #1475, opened 2021-12-15
Stop stiching uw

FredrikAreschoug:
# Jira Issue: [IPC-310]

## What?
- Stop stitching in underwriters GraphQL schema and resolve all queries and mutations here in giraffe.
- Mimic what underwriter did in the resolving of queries and mutations

## Why?
- Clearer separation of code.

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally(quite a lot :))
- [ ] Codescouted

[IPC-310]: https://hedvig.atlassian.net/browse/IPC-310?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/resolvers/headerUtils.ts
```diff
@@ -0,0 +1,14 @@
...
+  switch (headers['X-Platform']) {
+    case 'ANDROID':
+      return QuoteInitiatedFromDto.Android
+    case 'IOS':
+      return QuoteInitiatedFromDto.Ios
+    case 'WEB':
+    default:
+      return QuoteInitiatedFromDto.WebOnboarding
+  }
+}
```
bystam, line 14: This is only used in a quote-specific file. Doesn't it make sense to just place it as a private function in there rather than publicly exposed like this? It's not very reusable outside of quotes.

src/resolvers/quotes/quotes.ts
```diff
@@ -0,0 +1,245 @@
...
+    if (twoDigitYear > breakPoint) {
+      return `19${ssn}`
+    } else {
+      return `20${ssn}`
+    }
+  }
+  return ssn
+}
+
+const updateMemberContactInformation = async (quote: QuoteDto, memberService: MemberServiceClient) => {
```
bystam, line 217: I guess this is "finalize onboarding"? 😅

src/resolvers/sign-or-approve/signOrApprove.graphqls
```diff
@@ -0,0 +1,13 @@
+extend type Mutation {
+  signOrApproveQuotes(quoteIds: [ID!]!): SignOrApprove!
```
bystam, line 2: I first was surprised by this being here - wrapping things like this. But then I noticed it used to be a cross-schema-extension. This is nicer :)

src/schema.ts
```diff
@@ -176,10 +176,6 @@ const makeSchema = async (): Promise<SchemaIntrospectionResult> => {
       url: process.env.ACCOUNT_SERVICE_GRAPHQL_ENDPOINT,
       authorized: true,
     }),
-    introspectRemoteSchema(SchemaIdentifier.UNDERWRITER, {
-      url: process.env.UNDERWRITER_GRAPHQL_ENDPOINT,
-      authorized: false,
-    }),
```
bystam, line 182: 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥

bystam (approved):
Looks amazing! ^^ there are quite a few hard-to-grasp queries and mutations in here, but I think it's a real improvement that they end up visibly grouped on the surface like this rather than spread out in different places like before.

I had one small stylistic comment about a utils file 😅 but all in all amazing work!

src/api/upstreams/underwriter.ts
```diff
@@ -269,6 +263,182 @@ export interface MonetaryAmountV2Dto {
...
+interface ErrorResponseDto {
+  status:"error"
+  breachedUnderwritingGuidelines: BreachedGuideline[] | null;
+  errorCode: string;
+  errorMessage: string;
+}
+
+export interface GenericData {
+  insuranceType: string,
+  typeOfContract: string;
```
cpiehl1, line 289: Things like these should be picked up by prettier

cpiehl1 (approved):
Great stuff! 💯 There seems to be something wrong with your prettier though 🤔

src/resolvers/quotes/quotes.ts
FredrikAreschoug: Yeah this is "finalize onboarding" I don't like it and I really hope we can remove it!

src/api/upstreams/underwriter.ts
FredrikAreschoug: I will reinstall prettier and run it!

PR #1478, opened 2021-12-16
Use currency from quote carts market in campaign

tobiasbexelius:
# Jira Issue: [MX-422]

## What?
- This is a bit of a hack, but we don't have a way to easily support this properly atm.

## Why?
-

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [x] Codescouted

[MX-422]: https://hedvig.atlassian.net/browse/MX-422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


XSwifter (approved):

src/resolvers/quote-cart/quoteCart.ts
```diff
@@ -234,11 +234,15 @@ export const QuoteCartSubset: ResolverSubset = {
         return null
       }
 
+      const incentive = transformIncentive(res.incentive, session.market)
+
+      console.log(`INCENTIVE ${JSON.stringify(incentive)}`)
```
cpiehl1, line 239: Is this supposed to be here?

tobiasbexelius: NO, thanks 😄

src/resolvers/campaigns/campaigns.ts
```diff
@@ -18,18 +18,36 @@ export const CampaignSubset: ResolverSubset = {
         __typename: 'Campaign',
         ...res,
         ownerName: res.campaignOwner,
-        incentive: transformIncentive(res.incentive),
+        incentive: transformIncentive(res.incentive, null),
```
maksym-krupenko, line 21: I think using default param like market = null would be suitable here

maksym-krupenko (approved):

src/resolvers/campaigns/campaigns.ts
tobiasbexelius: Cool, did not know ts had that!

PR #1481, opened 2021-12-20
Connect payment without onboarding id

cpiehl1:
# Jira Issue: [MX-374]

## What?
- Update CPS API to no longer take `onboardingId`, and instead use `paymentTokenId`

## Why?
- It will be more consistent between logged in and logged out states

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [x] Codescouted

[MX-374]: https://hedvig.atlassian.net/browse/MX-374?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/resolvers/quote-cart/quoteCart.graphqls
```diff
@@ -62,13 +67,14 @@ as a means of storing intermediate state up until the point where it is "signed"
 into a proper "member".
 """
 type QuoteCart {
-    id: ID!,
+    id: ID!
+    market: Market!
```
styrelius, line 71: Does the `!` mean the `market` needs to be added to the `quoteCart` when it's created then?

package.json
```diff
@@ -38,6 +38,7 @@
     "graphql-tag": "^2.10.1",
     "graphql-tools": "^4.0.5",
     "ioredis": "^4.14.0",
+    "iterall": "^1.3.0",
```
styrelius, line 41: Couldn't find a file in this PR where this is used - do we need it?

cpiehl1: This was imported in `quoteCart.ts` but was only an implicit dependency, so I added it as an explicit dependency.

src/resolvers/quote-cart/quoteCart.graphqls
cpiehl1: This is already a required input in `CreateOnboardingCartInput`. Adding it in `type QuoteCart` simply means that you can always ask for it when querying the `QuoteCart`.

package.json
styrelius: I see 👌

src/resolvers/quote-cart/quoteCart.graphqls
styrelius: Aah, right! And before the changes in this PR we could get the market from the `QuoteCart` since we had the quote cart ID? But couldn't we still get the market from the quote cart - by using the Device ID header? Or isn't that the connection that we have to the quote cart? Or maybe it's better that it's an explicit input anyway 🤔

cpiehl1: Yea, we could get it from there. But as you said, I think it is better to have it explicit that in a header which is arguably not even required.

styrelius: True, forgot that the device id header will be optional 👍

styrelius (commented):
I think this looks reasonable 👌 But I also think someone with better insight into the bigger system should approve it, and not me 😳

src/api/upstreams/paymentService.ts
```diff
@@ -31,12 +30,11 @@ export const createPaymentServiceClient = (
...
   return {
-    connectPaymentDetails: async (onboardingId, request) => {
-      const suffix = onboardingId != null ? `?initiatedBy=${onboardingId}` : ''
+    connectPaymentDetails: async (request) => {
       const response = await client.post(
-        `${paymentDetailsUrlPrefix}${suffix}`,
+        `${paymentDetailsUrlPrefix}`,
         request,
-        { noAuthorization: onboardingId != null },
+        { noAuthorization: true },
```
tobiasbexelius, line 37: We should require token for logged in members right?

src/resolvers/quote-cart/quoteCart.ts
```diff
@@ -145,16 +143,37 @@ export const QuoteCartSubset: ResolverSubset = {
...
+      { upstream },
+    ): Promise<AddPaymentTokenResult> => {
+      const result = await upstream.underwriter.updateQuoteCart(id, {
+        paymentTokenId,
+      })
+
+      if (result == null) {
+        return transformSessionDto(await upstream.underwriter.getQuoteCart(id))
+      } else {
+        logger.info(
```
maksym-krupenko, line 158: I think 'error' level would be better to use here.

src/resolvers/quote-cart/quoteCart.ts
```diff
@@ -145,16 +143,37 @@ export const QuoteCartSubset: ResolverSubset = {
...
+          `Failed to add payment token with code: ${result.errorCode}`,
+        )
+        return {
+          __typename: 'BasicError',
+          message: 'Failed to add payment token',
```
maksym-krupenko, line 163: Maybe we should return some more informative message? Or it wouldn't be good from GDPR perspective?

hugokallstrom (commented):
Looks good to me, can't comment about the token though 🤔 Someone with more experience in api-gateway and payment-service should probably take a look as well.

maksym-krupenko (approved):

src/resolvers/quote-cart/quoteCart.ts
cpiehl1: Yea, pretty sure I read in Adyen's documentation that it is recommended to give as little information as possible to prevent fraud attempts

cpiehl1: Not sure, I just followed the pattern from adding campaigns

cpiehl1: But I think it might be good to not expose too much to the clients here, since we log the error anyway

PR #1486, opened 2022-01-05
POC of cross-sell and self-change quotes using QuoteCart API.

maksym-krupenko:
# Jira Issue: []

## What?
-

## Why?
-

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


tobiasbexelius (commented):
Neat, much less code than I thought it would be 😌

src/resolvers/quote-cart/quoteCart.ts
```diff
@@ -458,3 +477,40 @@ export const getQuoteCart = async (
...
+  input: CreateQuoteBundleInput,
+  upstream: Upstream,
+) => {
+  const contracts = await upstream.productPricing.getMemberContracts()
+  const activeHomeContentAgreement = getActiveHomeContentAgreement(contracts)
+  if (!activeHomeContentAgreement && !input.payload) {
+    throw new Error('There is no data to put in quote request')
+  }
+  const member = await upstream.memberService.getSelfMember()
+  const data = activeHomeContentAgreement || input.payload
```
cpiehl1, line 508: In the cross selling flow, we allow the member to specify number of co-insured. This means that it could be a different number from what is on their Home Content contract. Maybe we should instead here take the union of the input and the Home Content agreement, where we prioritise the input if a field is present in both?

cpiehl1 (approved):
Looks very crisp! 🔥

src/resolvers/quote-cart/quoteCart.graphqls
```diff
@@ -133,6 +133,13 @@ enum Market {
 
 input CreateQuoteBundleInput {
     payload: [JSON!]!
+    initiatedFrom: QuoteInitiatedFrom
+    contractBundleId: ID
```
cpiehl1, line 137: Where do the clients get the `contractBundleId` from?

maksym-krupenko: Can't say you for sure - it's used in existing cross-sell mutation so I added it in new API.

src/resolvers/quote-cart/quoteCart.ts
maksym-krupenko: Actually, I thought that numberCoInsured would be the field inside JSON payload.

src/resolvers/quote-cart/quoteCart.graphqls
tarikstafford: We get it from `activeContractBundles: [ContractBundle!]!` but we wouldn't have this outside of `CrossSell` right?

cpiehl1: I think it is used in moving flow, not cross-sell. But yes, should be only moving flow and nowhere else.

src/resolvers/quote-cart/quoteCart.ts
cpiehl1: But this statement means that if we have a non-null `activeHomeContentAgreement` we will ignore the payload, right?

src/resolvers/quote-cart/quoteCart.graphqls
maksym-krupenko: Yeah, it's optional field.

src/resolvers/quote-cart/quoteCart.ts
maksym-krupenko: yes, exactly

cpiehl1: Isn't that a problem if the payload contains a different `coInsured` than the `activeHomeContentAgreement`?

maksym-krupenko: Yeah, it's very valid point. Let me rework it.

PR #1487, opened 2022-01-07
Publish quote cart updates to subscribers after mutations

tobiasbexelius:
# Jira Issue: [MX-378]

## What?
- Previously updates were only published when signing with BankID (since that is the primary use case for the subscription, mutations already return the quoteCart). Now updates are published when the cart changes due to a mutation.

## Why?
- It makes more sense to deliver updates for all changes.

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [x] Codescouted

[MX-378]: https://hedvig.atlassian.net/browse/MX-378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bystam (approved):
Wow! Interesting approach to move both reading and writing of the redis stuff to Giraffe. Makes it was simpler to understand where things go in and out :D

PR #1492, opened 2022-01-20
Parse startSignResponse correctly

cpiehl1:
# Jira Issue: []

## What?
-

## Why?
-

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


src/resolvers/sign-or-approve/signOrApprove.ts
```diff
@@ -23,7 +24,7 @@ export const SignOrApproveSubset: ResolverSubset = {
           const res = await context.upstream.underwriter.signQuotes(quoteIds)
           return {
             __typename: 'SignQuoteResponse',
-            signResponse: res,
```
cpiehl1, line 26: Why did the compiler not catch this?

bystam (approved):

src/resolvers/sign-or-approve/signOrApprove.ts
bystam: Enough overlap? Something to do with Partial maybe?

PR #1493, opened 2022-01-21
Enable cross-sell in NO and DK in staging

cpiehl1:
# Jira Issue: []

## What?
-

## Why?
-

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


src/resolvers/uncategorised/crossSell.ts
```diff
@@ -41,42 +41,40 @@ export const crossSellResolver: Resolvers['CrossSell'] = {
...
 
-  // Temporarily disable cross-sell for NO and DK
-  if (contractTypes.length > 0 && !contractTypes[0].startsWith('SE')) {
-    return []
-  }
+    // Temporarily disable cross-sell for NO and DK
+    if (
+      contractTypes.length > 0 &&
+      !contractTypes[0].startsWith('SE') &&
+      process.env.SENTRY_ENV === 'production' // for some reason our node_env is 'production' in staging
```
cpiehl1, line 59: This line is the only real change

bystam: > for some reason our node_env is 'production' in staging

You're referencing `SENTRY_ENV` in the code, not `NODE_ENV`. Is that on purpose?

Also, I think the reason `NODE_ENV` is `production` in staging is because it's a reserved env variable used by [some node libraries](https://stackoverflow.com/questions/16978256/what-is-node-env-and-how-to-use-it-in-express), and I don't think it has a "staging environment" setting. It's simply meant to differentiate between a "local environment" (development) and a "deployed environment" (production).

cpiehl1: Yea I only want this change in staging so I need them to be different in staging and prod, `SENTRY_ENV` was the only good candidate I found for this by looking at the prod_env config

bystam (approved):

PR #1506, opened 2022-02-02
Fix/uw create quote fields

mightyost:
# Jira Issue: [IPC-461]

## What?
- Updated the Underwriter create quote request fields naming to be compatible with new generic.
  - The `data` field should carry the insurance data payload.
  - The boolean fields should be prefixed with `is`
  - The `householdSize` and `coInsured` should be `numberCoInsured`.

## Why?
- A prerequisite to migrate old insurances to generic.

## Optional checklist
- [x] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


src/resolvers/quote-bundle/quoteBundle.test.ts
```diff
@@ -127,9 +127,9 @@ const genericData: GenericData = {
   numberCoInsured: 0,
   extraBuildings: [],
   floor: null,
-  student: false,
-  subleted: false,
-  youth: false,
+  isStudent: false,
+  isSubleted: false,
+  isYouth: false,
```
mightyost, line 132: I hope it is fine (and more correct) to do these changes. UW should have returned them prefixed with `is` for quite a while. Does this mean this has been broken here?

In UW:
```
    @field:JsonProperty("isSubleted")
    val isSubleted: Boolean? = null,
    @field:JsonProperty("isStudent")
    val isStudent: Boolean? = null,
    @field:JsonProperty("isYouth")
    val isYouth: Boolean? = null
```

bystam: Probably 🤷 😅

bystam (approved):
LGTM

bystam:
> How can we make selfChange generic????

I don't know :) will probably be tricky

PR #1508, opened 2022-02-04
Adjust gradient map

sampettersson:


src/cross-schema/typeOfContractGradientOption.ts
```diff
@@ -20,14 +20,14 @@ const typeOfContractToGradient = (typeOfContract: string): string => {
...
+    'DK_HOME_CONTENT_OWN': two,
+    'DK_HOME_CONTENT_RENT': two,
+    'DK_HOME_CONTENT_STUDENT_OWN': two,
+    'DK_HOME_CONTENT_STUDENT_RENT': two,
     'DK_ACCIDENT': three,
     'DK_ACCIDENT_STUDENT': three,
-    'DK_TRAVEL': two,
-    'DK_TRAVEL_STUDENT': two
+    'DK_TRAVEL': one,
+    'DK_TRAVEL_STUDENT': one
```
StylianosGakis, line 30: This wasn't really specified in the previous implementation so I guess it was previously falling to the default case of `three`. So I guess this is another thing I don't know about 😅

StylianosGakis (approved):

PR #1509, opened 2022-02-07
Fixed more usages of old deprecated UW create quote fields

mightyost:
# Jira Issue: []

## What?
Found and fixed more usages of old (soon to be removed) UW create quote field namings.

## Why?
- To be able to migrate old insurances to generic.

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


cpiehl1 (approved):
🚀

src/resolvers/uncategorised/__snapshots__/crossSellQuotes.test.ts.snap
```diff
@@ -60,18 +60,18 @@ Array [
...
+        "isYouth": false,
         "lineOfBusinessName": "OWN",
         "livingSpace": 98,
         "numberCoInsured": 0,
         "postalCode": "12345",
         "squareMeters": 98,
         "status": "ACTIVE",
         "street": "Fakestreet 123",
-        "student": false,
+        "type": "DANISH_ACCIDENT",
```
sampettersson, line 73: [#L74](https://github.com/HedvigInsurance/giraffe/pull/1509/files#diff-750486e09738a49677fe5b1d6f54eea8d99b881cb0d8477557f9b5adb6a75d46R74) says DK_HOME_CONTENT_OWN, is that expected?

mightyost: Yes, saw that too. Its because a bunch of fields are dumped from the base for cross sell (`homeContentAgreement`)...

Not optimal but should not break anything since UW does not care about that field...

```
    const tasks = crossSellInsuranceTypes.map((insuranceType) => {
      const body: CreateQuoteDto = {
        ...member,
        // We are basically throwing in anything that underwriter
        // might care about and then it will pick out the necessary fields
        data: {
          ...homeContentAgreement,
          ...homeContentAgreement.address,
```

mightyost: Is it hard to clean out certain fields  from above in TypeScript?

cpiehl1: You could declare explicitly which fields to include. It will have to be the union of all the possible fields UW might ask for, so I guess it depends a bit on what all the schemas look like.

PR #1512, opened 2022-02-10
Fix X-Platform not always being defined

sampettersson:
## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [ ] Codescouted


FredrikAreschoug (approved):
Nice I just started to look at this, but I think you fixed it in a better way then me :smile:

PR #1514, opened 2022-02-10
Catch more errors when creating quotes

tobiasbexelius:
# Jira Issue: []

## What?
- Catch all somewhat probable error codes

## Why?
- If we try to create a bundle with more than one quote and the first quote is successfully created, and the other is not, we need to remove the first quote from the cart to not end up with a half created bundle. If we receive an error with a http status code that is not in the validStatusCodes list we will just crash and not get the opportunity to do so.

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [x] Codescouted


mightyost (approved):
👍

PR #1515, opened 2022-02-11
Fixed resolving of displayName on extra house

FredrikAreschoug:
:bug:


src/resolvers/utils.ts
```diff
@@ -319,12 +319,9 @@ export const transformExtraBuildingsDto = (
 ): ExtraBuildingValue[] => {
   let extraBuildings: ExtraBuilding[] = []
   if (buildings) {
-    if (!locale) {
-      locale = Locale.SvSe
-    }
     extraBuildings = buildings.map(
       (building) =>
-        transformExtraBuildingDto(building, strings) ?? throwOnNull(''),
+        transformExtraBuildingDto(building, strings.withLocale(locale ?? Locale.SvSe)),
```
mehdielket, line 324: It may be a bit better to default to English instead of Swedish?

XSwifter (approved):

PR #1523, opened 2022-03-07
PA-367 Add query get partner by id

alexander-rogachev-fastdev:
# Jira Issue: [PA-367]

## What?
- Add query get partner by id

## Why?
-

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted

[PA-367]: https://hedvig.atlassian.net/browse/PA-367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/resolvers/partners/partners.graphqls
```diff
@@ -0,0 +1,8 @@
+extend type Query {
+    partner(id: ID!): Partner!
+}
+
+type Partner {
```
sampettersson, line 5: Just a thought, this schema is used by a lot of teams at Hedvig, and Partner is a somewhat vague non descriptive term, maybe this type could be more explicit? like `RapioPartner` or similar.

src/resolvers/partners/partners.ts
```diff
@@ -0,0 +1,16 @@
+import { Partner } from '../../generated/graphql'
+import { ResolverSubset } from '../ResolverSubset'
+
+export const PartnerSubset: ResolverSubset = {
+  Query: {
+    partner: async (_parent, { id }, { upstream }): Promise<Partner> => {
+      if (!id) {
+        throw new Error('We need a id to fetch partner')
```
sampettersson, line 8: Generally you would type this if it's an expected outcome, like `union PartnerResponse = NotFound | Partner`. Also, you made ID explicitly non-nullable, it can never be null here, so your null check is unnecessary.

sampettersson (changes requested):

src/resolvers/partners/partners.ts
alexander-rogachev-fastdev: will fix

src/resolvers/partners/partners.graphqls
alexander-rogachev-fastdev: ok, np. will rename it.

sampettersson (approved):

PR #1535, opened 2022-03-14
translate extra building display name

FredrikAreschoug:
## What?
- translate extra building display name inside of json blob

## Why?
- because it was translated before

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


mightyost (approved):
👍

PR #1539, opened 2022-03-16
3-sec quote fixes after look up changes

FredrikAreschoug:
## What?
- 3-sec quote fixes after look up changes

## Why?
- :runner: 

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


robinandeer (approved):
Don't understand everything but would love to try it out - so: 👍

PR #1541, opened 2022-03-16
Add `bundleType` to quote

robinandeer:
# Jira Issue: []

## What?

Make it possible to differentiate main, addon, standalone quotes.
Move out inner function (`priority`).

## Why?

To support our "bundling" possibilities on the offer page.

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


src/resolvers/quote-bundle/quoteBundle.graphqls
```diff
@@ -13,6 +13,12 @@ type QuoteBundle {
   displayName(locale: Locale!): String!
 }
 
+enum BundledQuoteType {
```
hugokallstrom, line 16: Could we add some comments here to make it clear what the difference is between eg. main and standalone?

hugokallstrom (approved):

src/resolvers/quote-bundle/quoteBundle.graphqls
robinandeer: Ahh yes that's a very good point!

src/resolvers/quote-bundle/quoteBundle.graphqls
```diff
@@ -31,6 +37,7 @@ type BundledQuote {
   typeOfContract: TypeOfContract!
   initiatedFrom: String!
   displayName(locale: Locale!): String!
+  bundleType: BundledQuoteType!
```
guilhermespopolin, line 40: Nit:
Wouldn't something like `type` or `category` be a better name? `BundledQuote` it's about one specific quote and `bundleType` sounds like the type of a bundle.

guilhermespopolin (approved):

src/resolvers/quote-bundle/quoteBundle.graphqls
robinandeer: Sure, I also don't like the name but it's the best I had.

We already have a "type" field for the "insurance type" so don't want to use that.

Category also feels a bit too generic, no?

guilhermespopolin: Yeah, naming is hard 😞

PR #1542, opened 2022-03-21
Refactor/generic agreements

cpiehl1:
# Jira Issue: []

## What?
- Remove the cross-schema extensions for `contractPills` and `contractDetailsTable`
- Instead add these directly to `type Contract` in the local schema
- Create new GraphQL type `GenericAgreement`
- Add this to contract as `currentAgreementV2`
- Use generic agreement to resolve `contractPills` and `contractDetailsTable`

## Why?
- To reduce the time to implement a new promise in giraffe

Note: 60 % of the diff comes from test snapshots

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [x] Codescouted


src/cross-schema/detailstable/tableRenderer.ts
```diff
@@ -1,50 +1,92 @@
...
+          return this.strings('DETAILS_TABLE_SWEDISH_APARTMENT_TYPE_LABEL')
+        }
+        if (typeOfContract.startsWith('DK_HOME_CONTENT')) {
+          return this.strings('DETAILS_TABLE_DANISH_HOME_CONTENT_TYPE_LABEL')
+        }
+        if (typeOfContract.startsWith('NO_HOME_CONTENT')) {
+          return this.strings('DETAILS_TABLE_NORWEGIAN_HOME_CONTENT_TYPE_LABEL')
+        }
+        return null
+      },
```
cpiehl1, line 30: This is quite ugly, but it is basically the same thing we had before. We should probably figure out a scalable way to handle this. Maybe we can have the `ContractDetailsTable` live in PCMS?

cpiehl1:
One problem is that this change would cause an address pill and detail row to appear for accident insurance. Not sure how badly we want to avoid this. We could add some temporary logic to filter out undesired fields on certain contracts types while we figure out how this should be handled. This is essentially the same problem as in my comment above: How do we configure which fields should be used in different types of contracts?

src/cross-schema/detailstable/tableRenderer.ts
sampettersson: Yeah, we should definitely try to shove the table definition and structure into PCMS I think.

hugokallstrom:
> One problem is that this change would cause an address pill and detail row to appear for accident insurance. Not sure how badly we want to avoid this. We could add some temporary logic to filter out undesired fields on certain contract types while we figure out how this should be handled. This is essentially the same problem as in my comment above: It is a question of figuring out how to configure which fields should be used on different types of contracts.

Yeah good question. Could this also be handled from PCMS somehow? Perhaps using an enum for pills and tables - like CO_INSURED, ADDRESS etc. And then let giraffe populate with data from generic contract.

cpiehl1:
Yea, I like the idea of having an enum. Then we would have this split between lokalise and PCMS, which is probably not ideal in the long run. But as a first step I think it would be great because it would be really easy to implement.

PR #1543, opened 2022-03-28
Fix/edit quote

cpiehl1:
## What?
- Only send change set to UW edit endpoint.

## Why?
- UW edit supports change sets now, no need to first fetch the quote and patch it with changes in Giraffe.


src/resolvers/quote-cart/quoteCart.ts
```diff
@@ -199,7 +198,7 @@ export const QuoteCartSubset: ResolverSubset = {
           return {
             message: 'Failed to edit quote(s)',
             limits: result.breachedUnderwritingGuidelines,
-            type: payload.data.type || quote.data.type,
+            type: payload.data.type || 'Unknown',
```
cpiehl1, line 201: Feels like making an API call just to get the type of the quote is a bit unnecessary. Any caller should already know this property anyway.

cpiehl1: This class was probably created with the create operation in mind

mightyost (approved):
Lgtm

PR #1548, opened 2022-04-05
Remove householdSize parameter from cross-sell embark story names

SatishBV:
# Jira Issue: [APP-1516]

## What?
- Removes the `householdSize` parameter from the `embarkStoryName` that was added based on the `currentCoInsured`.

Note: I wasn't able to test this locally, but I did test that removing this parameter would actually solve the issue on iOS and Android. So, anyone more familiar with this codebase, kindly review the code and let me know if these changes are doing only the intended stuff of just removing the `householdSize` parameter and nothing else.

## Why?
- The price for any cross-sell was always getting calculated using the `householdSize` parameter present in the `embarkStoryName` for iOS. This was also complicating the messages being shown in embark for Android.
- Removing this parameter from the `embarkStoryName` would prevent the `householdSize` from being identified as a computed store value.

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted

[APP-1516]: https://hedvig.atlassian.net/browse/APP-1516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cpiehl1 (approved):
LGTM!

PR #1557, opened 2022-04-13
Fix/insurance terms date

cpiehl1:
# Jira Issue: []

## What?
- Use inception date when resolving terms documents
- Move a bunch of fields out of cross-schema extensions 

## Why?
- We need this to get the correct version
- These things can already be resolved locally so there is no reason to have them inside the cross-schema extension

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [x] Codescouted


hugokallstrom (approved):
LGTM

PR #1563, opened 2022-04-19
Add variations logic for Car

sampettersson:
TRAFFIC, HALF or FULL

Screenshot 2022-04-19 at 15 32 25

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [ ] Codescouted


src/api/upstreams/underwriter.ts
```diff
@@ -241,6 +241,7 @@ export interface GenericData {
   numberOfWetUnits: number | null
   isYouth: boolean | null
   zipCode: string | null
+  registrationNumber: string | null
```
AlexSJ1990, line 244: this won't be present on the GenericData - registrationNumber and mileage will be stored in a json blob called "data" on the QuoteDto and both will be mandatory.

src/resolvers/quote-bundle/quoteBundle.graphqls
```diff
@@ -77,6 +78,10 @@ type SwedishAccidentDetails {
   isStudent: Boolean!
 }
 
+type SwedishCarDetails {
+  registrationNumber: String!
```
AlexSJ1990, line 82: we should add `mileage: Int!` also

AlexSJ1990 (commented):
Looks nice! 2 small things

AlexSJ1990 (approved):
Sweeet!

PR #1564, opened 2022-04-19
add Swedish car

AlexSJ1990:
# Jira Issue: [IPC-577]

## What?
- Add Swedish Car to giraffe

## Why?
- to unblock clients

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted

[IPC-557]: https://hedvig.atlassian.net/browse/IPC-557?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[IPC-577]: https://hedvig.atlassian.net/browse/IPC-577?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/resolvers/quote-bundle/quoteBundle.graphqls
```diff
@@ -41,6 +41,7 @@ union QuoteDetails =
     SwedishApartmentQuoteDetails
   | SwedishHouseQuoteDetails
   | SwedishAccidentDetails
+  | SwedishCarAgreement
```
sampettersson, line 44: This should be `SwedishCarDetails` right?

src/cross-schema/contractPills.ts
```diff
@@ -225,6 +225,13 @@ export const createContractPills = (): CrossSchemaExtension => ({
               ... on SwedishAccidentAgreement {
                 numberCoInsured
               }
+              ... on SwedishCarAgreement {
+                address {
```
sampettersson, line 229: We probably don't want the address here, its more a system thing that the insurance has an address rather than a conscious user decision, so we tend to remove it on insurances where it doesn't make too much sense.

src/cross-schema/detailstable/contractDetailsFragments.ts
```diff
@@ -35,6 +35,17 @@ const agreementFragment = `
     coInsured: numberCoInsured
     activeFrom
   }
+  ... on SwedishCarAgreement {
+    address {
```
sampettersson, line 39: Same with the address here.

sampettersson (changes requested):

sampettersson (approved):
🎉

src/resolvers/contracts/contracts.ts
```diff
@@ -1,4 +1,4 @@
-import { format } from 'date-fns'
+import {format} from 'date-fns'
```
cpiehl1, line 1: Looks like your formatting is a bit different, do you have prettier installed?

cpiehl1 (approved):
Looks great!

src/resolvers/contracts/contracts.ts
AlexSJ1990: good point - have run prettier now :)

PR #1573, opened 2022-04-25
Car info details table

sampettersson:
# Jira Issue: [APP-1603]

[APP-1603]: https://hedvig.atlassian.net/browse/APP-1603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/resolvers/quote-bundle/quoteBundle.graphqls
```diff
@@ -78,9 +78,19 @@ type SwedishAccidentDetails {
   isStudent: Boolean!
 }
 
+type SwedishCarInfo {
+  model: String
+  makeAndModel: String
+  age: Int
+  color: String
+  ownerName: String
+  ownerSsn: String
```
AlexSJ1990, line 87: do we need this here? Guess we will never display to user..

src/api/upstreams/lookupService.ts
```diff
@@ -156,5 +169,23 @@ export const createLookupServiceClient = (
       const result = await response.json()
       return { status: 'success', ...result }
     },
+    async getSwedishCarInfo(
+      registrationNumber: string
+    ): Promise<SwedishCarResponse | null> {
+      const response = await client.get(
+        `/secure/lookup/car/${registrationNumber.replace(/ /g, "")}/info`,
```
simonauner, line 176: It's better to replace any whitespace character instead of just space, I guess

```suggestion
        `/secure/lookup/car/${registrationNumber.replace(/\s/g, "")}/info`,
```

PR #1574, opened 2022-04-26
Fix/replace non existing story name

hugokallstrom:
# Jira Issue: []

## What?
- Replace non-existing story names with update required-story

## Why?
- Some stories are not backwards compatible (with quote cart), we then show a message explaining that the member needs to update the app

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [ ] Codescouted


StylianosGakis (approved):

hugokallstrom:
Not needed since we are specifying story name in PCMS

PR #1579, opened 2022-04-29
IPC-617: adding endpoint for contract logo

adam-t-knight:
# Jira Issue: [IPC-617]

## What?
- Added getContractLogo

## Why?
- So that the app team can get qasa or hedvig logos stored in PCMS

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted

[IPC-617]: https://hedvig.atlassian.net/browse/IPC-617?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/resolvers/contracts/contracts.ts
```diff
@@ -499,5 +526,22 @@ export const transformTrialToFakeContract = (
...
+        dark: {
+          pdfUrl: '/uploads/Hedvig_19c5993150.pdf',
+          svgUrl: '/uploads/Hedvig_40448dd169.svg',
+          vectorDrawableUrl: '',
+        },
+      },
+      pdfUrl: '/uploads/Hedvig_19c5993150.pdf',
+      svgUrl: '/uploads/Hedvig_40448dd169.svg',
+      vectorDrawableUrl: '',
+    },
```
cpiehl1, line 545: I think you can leave this blank and the resolver you just implemented will take care of it

src/resolvers/contracts/contracts.ts
```diff
@@ -297,6 +307,23 @@ export const transformContract = (
...
```
cpiehl1, line 326: This one too

src/resolvers/contracts/contracts.ts
adam-t-knight: It seems to complain when I set it to null.

TS2322: Type 'null' is not assignable to type 'Icon'.

AmandaHedvigCirverius (approved):

src/resolvers/contracts/contracts.ts
cpiehl1: Ah right, this is because the codegen library doesn't use nested `Partial`. A bit unfortunate, but I guess we can leave it like this then

cpiehl1 (approved):

PR #1584, opened 2022-05-04
Add endpoint for getting FAQ based on TypeOfContract

simonauner:
# Jira Issue: [GRW-1029]

## What?
- Add Query for getting FAQ based on `TypeOfContract`. This is part of a larger data set as well, but we don't need all that data for the offer page so this PR breaks it out to a separate query.

## Why?
- We need to get Car-specific FAQ content in the offer page. Currently data is fetched from GraphCMS but we have most of the same data in PCMS also. Since PCMS can be queried by `ContractType` we can migrate the remaining data from GraphCMS to PCMS and use that for all FAQ:s.

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [x] Codescouted


robinandeer (commented):
Nice, to me it looks good but never worked with PCMS.

The only thing is the naming since we are fetching a list of FAQ items should we name it with that it mind?

- `getContractFaqs`
- `getContractFaqList`

Something like that?

simonauner:
> Nice, to me it looks good but never worked with PCMS.
> 
> The only thing is the naming since we are fetching a list of FAQ items should we name it with that it mind?
> 
> * `getContractFaqs`
> * `getContractFaqList`
> 
> Something like that?

Yes, I totally agree 👍

robinandeer:
@simonauner haha yeah okey it's a pretty established concept that an "FAQ" page has a bunch of questions/answers

Guess if we need to get a single item we can use `getContractFaqItem`

robinandeer (approved):
I'll approve the API design but someone who has worked with PCMS should approve the integration ;)

gustaveen (approved):
LGTM! Since it is an abbreviation for "frequently asked questions" (plural) I think it makes sense 👍

simonauner:
Anyone who works with PCMS who wants to approve as well? @cpiehl1

cpiehl1 (approved):
Nice! 💯

PR #1589, opened 2022-05-06
Get all cross-sell data from PCMS

cpiehl1:
# Jira Issue: []

## What?
- Get all cross-sell data from PCMS

## Why?
- This simplifies the required giraffe logic a lot

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [ ] Codescouted


hugokallstrom (approved):
Soooo good!!

src/resolvers/uncategorised/crossSell.ts
```diff
@@ -1,249 +1,69 @@
...
+      return {
+        type: crossSellData.productType as CrossSellType,
+        contractType: typeOfContract as TypeOfContract,
+        title: crossSellData.notificationTitle,
+        description: crossSellData.notificationDescription,
+        callToAction: crossSellData.notificationCta,
+        imageUrl: coverPhoto.coverPhotoUrl,
+        blurHash: coverPhoto.coverPhotoBlurHash,
+        info: {
+          displayName: strings(`CONTRACT_DISPLAY_NAME_${typeOfContract}`),
```
hugokallstrom, line 43: We should be able to remove this. @sampettersson are you using this field on iOS?

sampettersson: We seem to use this field on iOS actually, should we not?

hugokallstrom: We are using title, should be the same as contract name. Specified in PCMS :)

hugokallstrom: Can we deprecate `displayName` Carl?

cpiehl1: Sure!

PR #1593, opened 2022-05-09
Add quoteCart_selfChange mutation

cpiehl1:
# Jira Issue: []

## What?
- Add a `quoteCart_selfChange` mutation

## Why?
- Re-using `quoteCart_createQuoteBundle` for this was awkward since the logic is so different

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


hugokallstrom (approved):
Looks good, think its good if Bystam reviews as well though

src/resolvers/quote-cart/quoteCart.ts
```diff
@@ -136,6 +137,72 @@ export const QuoteCartSubset: ResolverSubset = {
...
+        upstream.productPricing.getMemberContracts(),
+      ])
+
+      const promises = contracts.map((contract) => {
+        const quoteData = {
+          ...member,
+          startDate: input.startDate,
+          data: {
+            ...input.payload,
+          },
```
bystam, line 156: Here we are kind of betting on the names of both the member object, all the different quote types and the input body to have overlapping names for the same concept. Is that something we can do now? When I built the original moving flow, there was quite a lot of fixing that had to be done.

But then again, worst case the quote creation just fails? Maybe not that bad.

src/resolvers/quote-cart/quoteCart.ts
```diff
@@ -136,6 +137,72 @@ export const QuoteCartSubset: ResolverSubset = {
...
+        },
+      )
+      if (updateQuotesResult !== null) {
+        logger.error(`failed to update quotes. Cart: ${id}`)
+        throw new InternalServerError()
+      }
+
+      const updatedSessionDto = await upstream.underwriter.getQuoteCart(id)
+      await updateSubscribers(pubsub, id)
+      return transformSessionDto(updatedSessionDto)
```
bystam, line 204: Looks like remnants of old naming. The word "session" I mean

bystam (approved):
LGTM, but I don't feel like I am an expert on this anymore 😅

src/resolvers/quote-cart/quoteCart.ts
cpiehl1: I feel like instead of hardcoding these field names in UW, giraffe and embark, maybe it is just simpler to only do it in UW and embark? The downside is that UW is further away from embark, so we might run a higher risk of only discovering errors at runtime. Maybe this is worth it though since we can avoid soooo much copy pasting. Especially now with pentagon, it has been so much easier adding new promises in the quoteCart parts of giraffe than the old cross-sell and moving-flow parts, since those parts aren't as generic.

PR #1594, opened 2022-05-10
Fix / Use display name as fallback for campaign display value

robinandeer:
# Jira Issue: [GRW-1080](https://hedvig.atlassian.net/browse/GRW-1080)

## What?
- Use a fallback for campaign display value if it fails for unexpected reasons
- Log the error

## Why?
- Giraffe fails when it can't properly translate a campaign display value

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [x] Codescouted


sampettersson (approved):

simonauner (approved):
Nice with a quick fix, but what is the actual problem?

Is it that `strings('GENERIC_CAMPAIGN_ADDEDPERK')` cannot be found?

FredrikAreschoug (approved):
:ok_hand:

robinandeer:
> Nice with a quick fix, but what is the actual problem?
> 
> Is it that `strings('GENERIC_CAMPAIGN_ADDEDPERK')` cannot be found?

It was a lokalise string that had been added with plural support. But we don't handle that in giraffe so it broke.

PR #1596, opened 2022-05-13
GRW-1029 / Add endpoint for FAQs

simonauner:
# Jira Issue: [GRW-1029]

## What?
- Add Redis to README
- Fix arg type for old GetContractFaq resolver
- Add new resolver for getting FAQ for multiple contracts at once

## Why?
For Car we need to make a change in the offer page where we get FAQ from PCMS (we used to get them from GraphCMS), since GraphCMS only has a generic FAQ and we want contract-type-specific FAQs for Car. When we make this change we need to get FAQ per contract type, and we can have several contract types in the offer page at the same time, so I would like to have an endpoint to query for multiple FAQs at the same time. 

Another slight complication is that we may have several contract types with the same FAQ content in PCMS, so I added a simple check to see if the headline content for one FAQ list is equal to that of another FAQ list, in which case we only return the first.

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [ ] Codescouted


src/resolvers/uncategorised/schema.graphqls
```diff
@@ -112,6 +116,25 @@ extend type Mutation {
...
+  contractTypes: [TypeOfContract!]!
+  """
+  Locale for PCMS
+  """
+  locale: Locale!
+}
+
+type ContractFaqResponse {
+  contractType: TypeOfContract
+  faqItems: [ContractFaq]
```
robinandeer, line 135: This will allow:

```
faqItems === null
```

and 

```
faqItems === [{...}, null, {...}]
```

Is that what we want or: 

```suggestion
  faqItems: [ContractFaq!]!
```

src/resolvers/uncategorised/schema.graphqls
```diff
@@ -33,6 +33,10 @@ extend type Query {
   """
   contractFaq(contractType: TypeOfContract!, locale: Locale!): [ContractFaq!]!
   """
+  Returns FAQ for TypeOfContract from promise-cms
+  """
+  contractFaqs(input: ContractFaqsQuery!): [ContractFaqResponse]!
```
robinandeer, line 38:
```suggestion
  contractFaqs(input: ContractFaqsQuery!): [ContractFaqResponse!]!
```

Or?

robinandeer (commented):
Nice update!

Just had some comments on schema definitions -- don't seem strict enough, or?

src/resolvers/uncategorised/schema.graphqls
simonauner: Yes, will fix both these comments! Thanks!

robinandeer (approved):
Looks good to me :)

PR #1601, opened 2022-05-17
Handle no/null response for getContractFaq from PCMS

simonauner:
# Jira Issue: [GRW-1029]

## What?
- Add error handling for when we cannot get ContractFaqs from PCMS so that we do not return 500

## Why?
- Now that PCMS can return 404 for ContractFaq when a Faq doesn't exist we need to handle it in Giraffe or it will return 500 https://github.com/HedvigInsurance/promise-cms/pull/102 

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [ ] Codescouted


src/resolvers/uncategorised/contractFaq.ts
```diff
@@ -8,4 +8,11 @@ export const contractFaq: QueryResolvers['contractFaq'] = async (
   _,
   args: QueryContractFaqArgs,
   { upstream },
-): Promise<ContractFaq[]> => upstream.promiseCms.getContractFaq(args)
+): Promise<ContractFaq[]> => {
+  try {
+    const result = await upstream.promiseCms.getContractFaq(args)
+    return result || []
```
robinandeer, line 14: nit: 
```suggestion
    return result ?? []
```

src/resolvers/uncategorised/contractFaq.ts
```diff
@@ -8,4 +8,11 @@ export const contractFaq: QueryResolvers['contractFaq'] = async (
...
+  } catch {
+    return []
```
robinandeer, line 16: Should we log some warning/error here that we can pick up in datadog?

robinandeer (approved):
Nice, and great with added tests :D

src/resolvers/uncategorised/contractFaq.ts
simonauner: It's quite expected that we get this error, but a log could be a nice way to detect when we have migrated all CMS content from GraphCMS - that means we can remove the fallback in web-onboarding when we get no more logs here. 👍

PR #1603, opened 2022-05-17
Car variant descriptions

sampettersson:
## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [x] Codescouted


simonauner (approved):
Without knowing much of the implementation details and implications: LGTM.

Don't know if my approval is enough or if someone else wants to chime in?

sampettersson:
I tested it locally with some quoteCarts from the web debugger and all looked fine, so and the tests seems to be happy, so i think its good to go :)

PR #1609, opened 2022-05-19
Map null fields from insurableLimits to empty string

sampettersson:
> This caused very weird errors 🙃 

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [ ] Codescouted


src/api/upstreams/promiseCms.ts
```diff
@@ -186,12 +186,22 @@ export const createPromiseCmsClient = (
...
-
-    return (await res.json()) || []
+    try {
+      const res = await client.get(
+        `/public/insurable-limits/${contractType}/${locale}`,
+        { optionalAuthorization: true },
+      )
+      const limits = (await res.json()) || []
+      
+      return limits.map((limit: InsurableLimit) => ({
```
simonauner, line 196: Will this mean that we can get a `limit` where all values are empty string? 👀

simonauner (approved):
gif

src/api/upstreams/promiseCms.ts
sampettersson: Yeah, but I don't think we need to do something in clients, someone in content will have to put it in, atleast things don't crash weirdly 😆

PR #1613, opened 2022-05-19
Add resolver for car info

sampettersson:
> This PR makes it possible to query for SwedishCarInfo without having an active quote or contract by just supplying the registrationNumber


AlexSJ1990 (approved):
Sweet!

simonauner:
Nice! 🙌

PR #1618, opened 2022-05-24
Remove 'Covers you' pill for Qasa contracts

SatishBV:
# Jira Issue: [APP-1686]

## What?
- Removes pill for `numberCoInsured` on Qasa contracts
- Sends the address pill even when the `numberCoInsured` pill is null

## Why?
- Implied that all Qasa contracts cover only one person. So this pill would not be necessary to be displayed.

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


sampettersson (approved):

cpiehl1 (approved):
🎉

PR #1619, opened 2022-05-24
Add trial info to init result

cpiehl1:
# Jira Issue: []

## What?
- Add trial info to the init widget result

## Why?
- Widget needs to know this

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


src/resolvers/partners/partners.graphqls
```diff
@@ -26,4 +26,10 @@ type PartnerInitWidgetResult {
     market: Market!,
     partnerName: String!,
     partnerDefaultCampaignCode: String
+    trialInfo: TrialInfo
+}
+
+type TrialInfo {
+    startDate: LocalDate!
+    endDate: LocalDate!
 }
```
cpiehl1, line 35: I think we should be able to ignore the trial id here, can we also ignore the move in date?

cpiehl1:
Lots of diff in the generated file, someone probably didn't run codegen before merging to master. But the only relevant changes here are in `partners.graphqls`.

src/resolvers/partners/partners.graphqls
Bjoernbom: Yea I guess! 👍🏼

Bjoernbom (approved):

Bjoernbom (approved):

bystam (approved):
Looks good to me

PR #1620, opened 2022-05-25
GRW-1029 / feat / Modify contractFaqs resolver to use ContractV2 endpoint in PCMS

simonauner:
# Jira Issue: [GRW-1029](https://hedvig.atlassian.net/browse/GRW-1029)

## What?
- Modify contractFaqs resolver to use ContractV2 endpoint in PCMS

## Why?
- This new endpoint returns more information about the FAQ, not just the FAQ items
   This allows us to show title of the FAQ, used to get "tabbable" FAQ section

This endpoint isn't used by anyone yet, so we can break it.

image

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [ ] Codescouted


src/resolvers/uncategorised/schema.graphqls
```diff
@@ -35,7 +35,7 @@ extend type Query {
   """
   Returns FAQ for TypeOfContract from promise-cms
   """
-  contractFaqs(input: ContractFaqsQuery!): [ContractFaqResponse!]!
+  contractFaqs(input: ContractFaqsQuery!): [ContractFaqV2!]!
```
robinandeer, line 38: Is there a point in returning an object with a nested list of `ContractFaqV2`?

I'm thinking that it would be easier to evolve the endpoint, no?

simonauner: Yes, the same thing I complained about in PCMS 🙃

simonauner: Fixed now! Let me know what you think

src/resolvers/uncategorised/schema.graphqls
```diff
@@ -35,7 +35,7 @@ extend type Query {
   """
   Returns FAQ for TypeOfContract from promise-cms
   """
-  contractFaqs(input: ContractFaqsQuery!): [ContractFaqResponse!]!
+  contractFaqs(input: ContractFaqsQuery!): ContractFaqsResponse!
```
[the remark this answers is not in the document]
robinandeer: I think this query was introduced to support the Car offer page on web so @simonauner should be on top of that I believe ;)

robinandeer (approved):
I live the new API. Allows us to add things like pagination meta data without breaking the API in the future :)

src/resolvers/uncategorised/schema.graphqls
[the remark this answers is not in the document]
simonauner: @oscarnylander Yes, @robinandeer is correct. This is a new endpoint that's not yet used in production, no need to sweat 🙃

PR #1621, opened 2022-05-25
Re-use subtype for non-home content contracts in moving flow

cpiehl1:
# Jira Issue: []

## What?
- When creating moving flow quotes, for non-home content we need to re-use the type and subType from the existing agreement. We were only doing this for type, which works for accident and travel but not car.

## Why?
-

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


rasmusguterstam (approved):
🚀

PR #1625, opened 2022-05-30
Filter out terminated contracts in moving flow

cpiehl1:
# Jira Issue: []

## What?
-

## Why?
-

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


src/resolvers/quote-cart/quoteCart.ts
```diff
@@ -147,7 +148,13 @@ export const QuoteCartSubset: ResolverSubset = {
...
       ])
 
-      const promises = contracts.map((contract) => {
+      const tasks = contracts.filter((c) =>
+        [ContractStatusDto.ACTIVE, ContractStatusDto.ACTIVE_IN_FUTURE].includes(
+          c.status,
+        ),
+      )
+
+      const promises = tasks.map((contract) => {
```
StylianosGakis, line 157: a nit probably, but not sure about the `tasks` name here, maybe I'm missing some context as to what it may mean.

I'd just chain the filter before the map we were doing before anyway without introducing a new variable probably. Does tasks mean something else here?

PR #1634, opened 2022-06-08
Return request id when initializing widget

cpiehl1:
# Jira Issue: []

## What?
-

## Why?
-

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


Bjoernbom (approved):
Looking good! I don't know about the asynchronous rapio call if it still needs to `await`, but I guess you know that better than I do. 👍🏼

cpiehl1:
Pretty sure all resolvers are async anyway so should be fine!

PR #1637, opened 2022-06-09
GRW-1196 Fix sanitised values passed to backend

mehdielket:
# Jira Issue: []

## What?
-

## Why?
-

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


simonauner:
So, the previous PR https://github.com/HedvigInsurance/giraffe/pull/1632 seems to have broken all GraphQL requests since the response object was directly modified by the sanitizer.

This PR changes that so that the sanitizer returns a sanitized copy of the object for logging purposes instead.

guilhermespopolin (approved):
The classic mutable/immutable problem

PR #1653, opened 2022-07-12
Fix / Survive errors when PCMS has missing data

robinandeer:
# Jira Issue: [GRW-1241](https://hedvig.atlassian.net/browse/GRW-1241)

## What?
- Add error handling when there's missing data in PCMS
- Log errors to datadog but return empty array to API

## Why?
- Hard to figure out why queries are failing
- We should setup alarms for this in Datadog

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [x] Codescouted


AlexSJ1990 (approved):
✅

src/api/upstreams/promiseCms.ts
```diff
@@ -201,22 +210,33 @@ export const createPromiseCmsClient = (
...
-        description: limit.description || "",
-        type: limit.type || ""
+        label: limit.label || '',
+        limit: limit.limit || '',
+        description: limit.description || '',
+        type: limit.type || '',
       }))
-    } catch (e) {
+    } catch (error) {
+      logger.error(`PCMS - Failed to get insurable limits for ${contractType}`)
```
AlexSJ1990, line 221: if we amend this to be #IPC-alert before the PCMS - we should get an alarm when this happens in prod

robinandeer: aha yeah that's a great feature -- I will add it :)

AlexSJ1990 (approved):

PR #1655, opened 2022-07-18
Fix issue with attributedTo to default to hedvig

johan-hg:
# Jira Issue: []

## What?
- When a previous quote does not exist we default `quotingPartner` to null causing Unterwriter to assume Hedvig.

## Why?
- In an attempt to fix a previous issue, we started defaulting to null.

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


cpiehl1 (approved):

robinandeer (approved):
Nice catch :)

PR #1656, opened 2022-08-10
IPC-751: initial updates for PCMS-2

adam-t-knight:
# Jira Issue: [IPC-751]

## What?
- Making Giraffe not return an error if PCMS returns null data. Also beginning updates for converting calls to PCMS to use a version string.

## Why?
- Since data has been restructured to be by Terms and Conditions instead of by Contract, some data previously obtained through Contracts will be missing. Currently this generates an error in Giraffe. This should hopefully at least provide empty data for the time being until everything can be converted to use a version string and go through Terms and Conditions.

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [ ] Codescouted

[IPC-751]: https://hedvig.atlassian.net/browse/IPC-751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cpiehl1 (approved):
LGTM

PR #1658, opened 2022-08-16
feat(QuoteBundle): added standaloneQuotes/additionalQuotes fields

guilhermespopolin:
## What?

- Adds `standaloneQuotes`/`additionalQuotes` fields to `QuoteBundle`

## Why?

- In order to support _Pentagon House Cross Sell_ we need to update the API so it implements the concept of _Standalone_/_Additional_ quotes. It can't be solved with `possibleVariations` because rather them let users pick a defined set of quotes (a variant) we need to allow them _build_ their combination while in the _Offer Page_

**A lot of the changes in this PR are just about prettier changes. The relevant part is the changes on [`getQuoteBundle`](https://github.com/HedvigInsurance/giraffe/pull/1658/files#diff-ce55f82ba87c3893bffd5f86719287bbab980dc90899417e1338188d2fc0cbf7R206) function.**

image

Jira Issue: [GRW-1220](https://hedvig.atlassian.net/browse/GRW-1220)


simonauner (approved):
So nothing is breaking in terms of API contract, right? It just adds `standaloneQuotes`/ `additionalQuotes`, and `standaloneQuotes` means something like "main quote"?

Looks good 👍

PR #1663, opened 2022-08-22
fix(quoteBundle): standalone/additional quotes possible variation calculation

guilhermespopolin:
## What?

- Fix a bug in the logic that calculates possible variations for NO/DK based on standalone/additional quotes

## Why?

- Not all possible variations were being listed

[Jira ticket](https://hedvig.atlassian.net/browse/GRW-1337)


robinandeer (approved):
hard to read but...

PR #1666, opened 2022-08-24
Remove key-gear

schmee-hg:
# Jira Issue: []

## What?
- key-gear has been sunset.

## Why?
- We're not using it or developing it.

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


src/resolvers/uncategorised/member.ts
```diff
@@ -41,41 +41,11 @@ export const memberFeatures: MemberResolvers['features'] = async (
...
-    throw new Error('failed to fetch member features')
-  }
-
-  const features = result.data.userFeatures
-    .filter((feature: { feature: string }) => feature.feature === 'KeyGear')
-    .map((_: { feature: string }) => Feature.KeyGear)
-
   const eligibleForReferrals = await isEligibleForReferrals(token, headers)
-
+  const features: Array<Feature> = []
```
schmee-hg, line 45: Is this even TS syntax? 🙈 Sorry I'm clueless about TS 😂

cpiehl1: It is correct syntax but I think it would be preferred to use `Feature[]` over `Array<Feature>`. Maybe we can just do something like this though:

`return eligibleForReferrals ? [Feature.Referrals] : []`

cpiehl1 (approved):
LGTM!

src/resolvers/uncategorised/member.ts
schmee-hg: Updated, thanks!

